PR #8, opened 2023-04-05
Configure Cypress for local UI testing

### LINKED ISSUE #5: Add a Testing Framework to prevent Regressions in Features
We should have E2E (end-to-end) tests included for each page and user flow across our website to ensure that more serious features do not see regressions and new bugs accidentally introduced by PRs as we continue working.

We should set up Cypress in a testing folder for an easy library structure that mimics tests on the portal backend.
https://github.com/vercel/next.js/blob/canary/examples/with-cypress/cypress/e2e/app.cy.ts

farisashai:
Configured Cypress the our project as our testing framework (closes #5). See `cypress/e2e/pages/auth/login.cy.ts` for an example test passing.

Added two new start scripts `yarn test` to run Cypress all tests through the CLI and `yarn test:ui` to open the Cypress window and let you visualize your tests and debug any failing output.

Example of Cypress UI output running through each line of test code.

Screenshot 2023-04-05 at 1 48 04 PM


raymosun (approved):
Looks good. This is so cool! Just wondering if we wanted to have the test scripts start the server as well?

farisashai:
> Looks good. This is so cool! Just wondering if we wanted to have the test scripts start the server as well?

So I was considering this and I think it makes the most sense to keep it separate because we will typically be running the website in development and later want to test our changes on the current running localhost website.

- If we have to start up the entire dev server on localhost every time we want to run the test and make any tweaks, it'll take way longer to rebuild and lint the full repo and then run tests afterwards if we tweak the test code repeatedly.
- If the test script also starts the server while it was previously running separately, it won't start on the right port for the testing config to pick up the correct website.

alexzhang1618 (approved):
Changes look good to me 😎

cypress/e2e/pages/auth/login.cy.ts
```diff
@@ -0,0 +1,20 @@
+/// <reference types="cypress" />
+
+describe('Login Page', () => {
+  beforeEach(() => {
+    cy.visit('/login');
```
trevorkw7, line 5: Not super familiar with sequence in which tests occur in Cypress when there are multiple tests but I could imagine that if this test is run after another test that requires login it would fail because we aren't logging the user out first before the test. Should we be ensuring that the user is logged out first here with a `cy.visit('/logout')`?

trevorkw7:
Other than considering this, we might also want to update the `README.md` to document the new `yarn test` and `yarn test:ui` commands in this section:
Screen Shot 2023-04-07 at 12 07 58 AM

trevorkw7:
Other than those suggestions, I think this looks good to go!

PR #12, opened 2023-04-06
Implemented page redirection for logged out users

### LINKED ISSUE #1: Enable page redirection to work properly
Currently on the live membership portal, if you visit a page such as `members.acmucsd.com/store` directly from a URL without being logged in, you will be send to the login page with an additional query parameter (e.g. `https://members.acmucsd.com/login?destination=/store`) allowing you to easily recover your intended place in the application. With Next.js, we can recreate this flow a lot simpler with only two files to change.

1. Generating the query parameter correctly in `src/lib/hoc/withAccessType` that handles authentication for all pages in our application.
2. Reading the destination page route from the query parameter (if it exists) correctly on the login page server-side and using that as the location to send the user once they are logged in correctly.

If you want to tackle additional fine details, 
1. add a toast icon letting them know why they were redirected ("You need to be logged in to view this page" if no user cookie, "You don't have permission to view this page" if their access type is incorrect).
2. add a third argument to the `withAccessType` function with a default value of login that lets you specify where to redirect them if they fail the access check.

trevorkw7:
- Removed `trailingSlash` from Next.js redirect urls
- Modified login page to redirect to read query parameters to log users in directly to originally requested page to resolve issue #1 
- Modified `src/lib/hoc/withAccessType` to generate query parameters for users who attempt to reach a page that requires login

### DIFF
--- a/src/lib/hoc/withAccessType.ts
+++ b/src/lib/hoc/withAccessType.ts
@@ -24,7 +24,7 @@ export default function withAccessType(
     const originalReturnValue = await gssp(context);
 
     // Save current URL so user can return here after logging in
-    const currentUrl = encodeURIComponent(req.url ?? '/');
+    const currentUrl = encodeURIComponent(req.url ?? config.homeRoute);
 
     const baseRoute = redirectTo ?? config.loginRoute;
     const destinationRoute =
@@ -39,9 +39,10 @@ export default function withAccessType(
 
     // Get user cookie
     const userCookie = CookieService.getServerCookie(CookieType.USER, { req, res });
+    const token = CookieService.getServerCookie(CookieType.ACCESS_TOKEN, { req, res });
 
     // If cookie is misisng, we are not logged in so don't show the current page
-    if (!userCookie) {
+    if (!userCookie || !token) {
       if (currentUrl === config.homeRoute) {
         return {
           redirect: {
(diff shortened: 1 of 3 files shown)


src/lib/hoc/withAccessType.ts
```diff
@@ -21,12 +21,14 @@ export default function withAccessType(
   const modified: GetServerSideProps = async (context: GetServerSidePropsContext) => {
     // Initialize variables
     const { req, res } = context;
-
     const originalReturnValue = await gssp(context);
-    // TODO: Save current URL so user can return here after logging in
+
+    // Save current URL so user can return here after logging in
+    const currentUrl = encodeURIComponent(req.url ?? '/');
```
farisashai: This magic variable should be `config.homeRoute`

src/lib/hoc/withAccessType.ts
```diff
@@ -35,7 +41,17 @@ export default function withAccessType(
     const userCookie = CookieService.getServerCookie(CookieType.USER, { req, res });
 
     // If cookie is misisng, we are not logged in so don't show the current page
-    if (!userCookie) return safeRedirectToLogin;
+    if (!userCookie) {
```
farisashai: or if auth token is missing, we shouldn't log in

farisashai (changes requested):
fix

farisashai (approved):
Merge after checks run

PR #17, opened 2023-04-09
Add Validation Service

farisashai:
This pull request introduces a new ValidationService class to handle all logic of validating input fields. 

While some parts of the code are not any shorter than previously, they are now managed from a single source of truth to allow for easy management in the future if any logic is to change **(e.g. valid password must now be 10 characters or fully alphanumeric)**


trevorkw7 (approved):
Looks great! Tested functionality on email validation on the Forgot Password and Login pages. The the use of short circuiting and success and error type definitions also looks like a really smart way to decrease code clutter.

PR #20, opened 2023-04-09
Refactored all classes into modules

farisashai:
- Typed all interface type imports correctly
- Exported all utility functions directly instead of in class objects to provide better tree-shaking, code exfoliation warnings for dead code, cyclic dependencies, and allow for code splitting of individual functions
- Export all `src/lib` and `src/components` functions from an `index.ts` file in the root of the project to enforce export naming conventions (default exports allow for renaming to anything which is more typo-prone) and allows us to name all module exports using `import * as MyService` to keep the remainder of our codebase fully intact without modifying any usage of the functions.


alexzhang1618 (dismissed):
Looks good to me

alexzhang1618 (approved):
Note: Added `position: fixed` to navbar so the navbar will stay in place for mobile (was previously `position: absolute`

PR #28, opened 2023-05-08
Implemented New Mobile Navbar UI

### LINKED ISSUE #3: Mobile Navbar Redesign
As part of the Store redesign, we will be simplifying the nav menu to include a relevant set of top level pages to remove the need for a sidebar or mobile submenu.

Screenshot 2023-04-04 at 5 29 34 AM

Screenshot 2023-04-04 at 5 30 35 AM

trevorkw7:
# Info

Closes **#3**. 

# Description

- Implements the mobile navbar dropdown UI on devices with width smaller than 768px

## Changes

- Added `vars.module.scss` to the navbar component to bounce global breakpoint vars into the navbar `index.tsx`
- Created a useWindowSize() function that attaches an event listener to the window and uses useState and useEffect to report the current window size of the UI
- Modified CSS and JSX of the Navbar component to hide the portal links and icon links on mobile devices in place of a new dropdown menu

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.
- [x] I have run and passed all new and existing Cypress tests. Add screenshots below.

# Checklist

      anywhere else.

# Screenshots

Dark Mode Menu Closed
image

Dark Mode Menu Opened
image

Light Mode menu Closed
image

Light Mode Menu Opened
image

Cypress Tests Passed

image


src/lib/utils.ts
```diff
@@ -27,3 +28,28 @@ export const getMessagesFromError = (errBody: CustomErrorBody): string[] => {
...
   return errBody.errors.map(err => getAllErrMessages(err)).flat();
 };
+
+// Define general type for useWindowSize hook, which includes width and height
+export interface Size {
+  width: number | undefined;
+  height: number | undefined;
+}
+
+export function useWindowSize(): Size {
```
farisashai: Let's add a JSDoc comment explaining the usage of this function.

farisashai: Since this is a specifically a custom React hook rather than a basic utility function, maybe we should store it in a file `src/lib/hooks/useWindowSize.ts`

src/lib/utils.ts
```diff
@@ -27,3 +28,28 @@ export const getMessagesFromError = (errBody: CustomErrorBody): string[] => {
...
+  });
+  useEffect(() => {
+    function handleResize() {
+      setWindowSize({
+        width: window.innerWidth,
+        height: window.innerHeight,
+      });
+    }
+    window.addEventListener('resize', handleResize);
+    handleResize();
```
farisashai: I think that if we initialize the `windowSize` useState with window.innerWidth and window.innerHeight, we can avoid having to call this function here and also remove the possible `undefined` type.

i'm not sure if this would throw an error with server-side rendering of the page if window is undefined though so you might need to check that

src/components/layout/Navbar/vars.module.scss
```diff
@@ -0,0 +1,8 @@
+/* stylelint-disable property-no-unknown */
+@use 'src/styles/vars.scss' as vars;
+
+:export {
+  breakpointLg: vars.$breakpoint-lg;
+  breakpointMd: vars.$breakpoint-md;
+  breakpointSm: vars.$breakpoint-sm;
+}
```
farisashai: I really appreciate being thoughtful to avoid re-defining the same variable values but like we mentioned in the discussion on the Discord, I can see a few issues with this approach.

1. You re-exported all of the values specifically for this folder to use in the Navbar. If we need to use CSS variables in a different JS component and create a file like this again to access them in every single folder, I think that creates a messier codebase with redundant duplication than storing these variables at a global level somewhere in the `lib` folder.
2. Assuming we move these variables and import then from `src/lib` instead, I somewhat prefer writing the variables in JS than pulling from CSS because we only have to rewrite them a single time and we have the guarantee that their values exist. For example, you had a line elsewhere where you wrote `parseInt(vars.breakpointMd || '768')` and the OR variable assignment shouldn't be necessary because this is a constant rather than a dynamically set value.

A simple approach I can see working is just having

`src/lib/config.ts`

```
const config = {
  cssVars: {
    breakpointMd: 768
  }
}
```
This gives us the benefit of a second but still single source of truth for all CSS values we need to use in JS as long as we keep them synchronized, along with a guarantee that the variable exists as the exact integer we need, no additional files to generate, and we can easily reference the variable's source and value if we look at the code.

src/components/layout/Navbar/style.module.scss
```diff
@@ -77,7 +89,86 @@
     background: vars.$wainbow;
     height: 0.25rem;
     position: absolute;
-    top: 4rem;
+    top: 4rem; // Position the wainbow at the bottom
+    transform: translateY(0); // Move wainbow to bottom of navbar by default
```
farisashai: is this line necessary?

src/components/layout/Navbar/style.module.scss
```diff
@@ -77,7 +89,86 @@
...
+    transition: 0.3s ease-in-out; // Apply transition to transform property
     width: 100vw;
   }
+
+  .mobileNav {
+    align-items: right;
```
farisashai: Should be `flex-end` instead of `right`

valid properties for this are `flex-start`, `center`, and `flex-end`, 
Screenshot 2023-05-08 at 10 05 24 AM

src/components/layout/Navbar/style.module.scss
```diff
@@ -77,7 +89,86 @@
     background: vars.$wainbow;
     height: 0.25rem;
     position: absolute;
-    top: 4rem;
+    top: 4rem; // Position the wainbow at the bottom
+    transform: translateY(0); // Move wainbow to bottom of navbar by default
+    transition: 0.3s ease-in-out; // Apply transition to transform property
```
farisashai:
```suggestion
    transition: 0.3s ease-in-out transform; // Apply transition to transform property
```
In your comment you mentioned you want to apply transitions to the transform property but we didn't actually specific that anywhere. By default, it will transition all properties on the object if you don't clarify so I like to mention this explicitly usually.

Screenshot 2023-05-08 at 10 08 42 AM

src/components/layout/Navbar/index.tsx
```diff
@@ -1,18 +1,39 @@
...
+import ACMIcon from '@/public/assets/icons/acm-icon.svg';
+import CalendarIcon from '@/public/assets/icons/calendar-icon.svg';
+import LeaderboardIcon from '@/public/assets/icons/leaderboard-icon.svg';
 import ProfileIcon from '@/public/assets/icons/profile-icon.svg';
 import ShopIcon from '@/public/assets/icons/shop-icon.svg';
 import Image from 'next/image';
 import Link from 'next/link';
-import { memo } from 'react';
+import { memo, useEffect, useState } from 'react';
+import { useWindowSize } from '../../../lib/utils';
```
farisashai: nit: absolute imports (for better auto-fixing of imports when we move files and avoiding lots of confusing `../`) for anything that's not in the same folder already

src/components/layout/Navbar/index.tsx
```diff
@@ -1,18 +1,39 @@
...
 const Navbar = ({ user }: NavbarProps) => {
+  const size = useWindowSize();
+
+  const [menuOpen, setMenuOpen] = useState(false);
+  const [mobile, setMobile] = useState(false);
+  const toggleMenu = () => setMenuOpen(!menuOpen);
+
+  // Switch to mobile less than breakpointMd
+  useEffect(() => {
+    setMobile((size.width || 0) <= parseInt(vars.breakpointMd || '768', 10));
```
farisashai: I mentioned this where you defined the hook but if we can remove the `undefined` from the type signature on `size.width` and initialize it correctly, I think we can avoid needing this default value.

farisashai: Also, be extra careful with assigning a default value this way since `0` is still a falsy value. If you're trying to assign a default value specifically when no value exists (`size.width === undefined`), you should use nullish coalescing (`??`) instead for a stronger condition

src/components/layout/Navbar/index.tsx
```diff
@@ -1,18 +1,39 @@
...
 
 interface NavbarProps {
   user: PrivateProfile;
 }
 const Navbar = ({ user }: NavbarProps) => {
+  const size = useWindowSize();
+
+  const [menuOpen, setMenuOpen] = useState(false);
+  const [mobile, setMobile] = useState(false);
+  const toggleMenu = () => setMenuOpen(!menuOpen);
```
farisashai: nit: let's use a callback function in the setState to avoid possible bugs here

src/components/layout/Navbar/index.tsx
```diff
@@ -1,18 +1,39 @@
...
+import vars from './vars.module.scss';
 
 interface NavbarProps {
   user: PrivateProfile;
 }
 const Navbar = ({ user }: NavbarProps) => {
+  const size = useWindowSize();
+
+  const [menuOpen, setMenuOpen] = useState(false);
+  const [mobile, setMobile] = useState(false);
```
farisashai: Let's name this variable a little bit more descriptively to something like `isMobile` or `isMobileView`

src/components/layout/Navbar/index.tsx
```diff
@@ -30,17 +51,27 @@ const Navbar = ({ user }: NavbarProps) => {
...
     <header className={styles.header}>
       <div className={styles.content}>
-        <Link href={config.homeRoute} className={styles.icon}>
+        {/* Mobile Navbar Toggle */}
+        <button
+          type="button"
+          className={`${styles.toggleIcon} ${!mobile && styles.hidden}`}
+          onClick={toggleMenu}
+        >
+          <div className={`${styles.bar1} ${menuOpen && styles.open}`} />
```
farisashai: This is something I learned recently and it doesn't change anything functionally tbh but I would prefer using data attributes like we did in the FE interview takehome to indicate the current state of an element because a data-attribute represents a changing property of this tag, rather than a static CSS style and it allows for easy selection and identification of the current state without extra class names.

This would look something like

`<div className={styles.bar1} data-open={menuOpen}>`

```
.bar1[data-open="true"] {
}
```

src/components/layout/Navbar/index.tsx
```diff
@@ -30,17 +51,27 @@ const Navbar = ({ user }: NavbarProps) => {
   return (
     <header className={styles.header}>
       <div className={styles.content}>
-        <Link href={config.homeRoute} className={styles.icon}>
+        {/* Mobile Navbar Toggle */}
+        <button
+          type="button"
+          className={`${styles.toggleIcon} ${!mobile && styles.hidden}`}
```
farisashai: Same thing as the comment below on data-attributes but we could describe this as something like
`
<button type="button" className={styles.toggleIcon} data-state={mobile ? "visible" : "hidden"}>
`
```
.toggleIcon[data-state="visible"] {
}
```

Also, I would also flip from calling it hidden to visible if I was keeping the current behavior to avoid a extra negation on the `mobile` boolean and make the code easy to comprehend for anyone reading.

src/components/layout/Navbar/index.tsx
farisashai: Also, unless this is being used in multiple places in the code, single line arrow functions can be inlined directly where they're used.

src/components/layout/Navbar/index.tsx
```diff
@@ -50,6 +81,34 @@ const Navbar = ({ user }: NavbarProps) => {
...
+        <Link className={styles.mobileNavItem} onClick={() => setMenuOpen(false)} href="/">
+          <CalendarIcon color="var(--theme-text-on-background-1)" className={styles.iconLink} />
+          Events
+        </Link>
+        <Link
+          className={styles.mobileNavItem}
+          onClick={() => setMenuOpen(false)}
+          href="/leaderboard"
+        >
+          <LeaderboardIcon color="var(--theme-text-on-background-1)" className={styles.iconLink} />
```
farisashai: Is there a reason we're defining the color here instead of in the `.iconLink` class name?

src/components/layout/Navbar/index.tsx
```diff
@@ -50,6 +81,34 @@ const Navbar = ({ user }: NavbarProps) => {
           </Link>
         </nav>
       </div>
+      {/* Mobile Menu Dropdown */}
+      <div className={`${styles.mobileNav} ${menuOpen && styles.open}`}>
+        <Link className={styles.mobileNavItem} onClick={() => setMenuOpen(false)} href="/">
```
farisashai:
```suggestion
        <Link className={styles.mobileNavItem} onClick={() => setMenuOpen(false)} href={config.homeRoute} >
```
Let's try to avoid magic variables for important routes (home, login, admin) to keep code readable as to what it's doing at a high level and make modifications easy if we ever needed to move the page route.

src/components/layout/Navbar/style.module.scss
```diff
@@ -48,6 +52,10 @@
       gap: 1rem;
       line-height: 24px;
       margin-right: auto;
+
+      &.hidden {
```
farisashai: Instead of adding `.hidden` to every element on the navbar, it might be easier to track the state of the whole navbar and use a single selector on the entire navbar to select everything at once that you need to change.

first line is bad psuedo-code
```
.header[mobile] {
  .iconLink {
    display: none;
  }
  .portalLinks {
    display: none;
  }
  etc
}
```

src/components/layout/Navbar/style.module.scss
```diff
@@ -77,7 +89,86 @@
...
+      border-radius: 2px;
+      height: 5px;
+      position: absolute;
+      top: 1.5px;
+      transition: 0.3s ease-in-out all;
+      width: 40px;
+
+      &.open {
+        transform: rotate(-45deg) translateY(-10px);
+        transform-origin: top right;
```
farisashai: This is kind of hard to describe just in words but the reason that transform-origin was top right on the main website is because this is the top bar and it was right aligned on the page so this line let us keep the top right corner fixed in place. Since our navbar here is left-aligned, we want to keep the top left corner fixed in place so the bars are basically just rotating instead of moving.

Screenshot 2023-05-08 at 10 35 45 AM

src/components/layout/Navbar/style.module.scss.d.ts
```diff
@@ -1,4 +1,6 @@
 export type Styles = {
+  mobileNavItem: string | undefined;
```
farisashai: is this the auto-generated code? I don't think the class names can be undefined since they're coming from the scss file. (I couldn't figure out a good way to generate these types on save in the codebase so it generates them automatically whenever you restart and run `yarn dev`. stretch goal to fix that 👀 )

farisashai (changes requested):
Good progress so far and is starting to come together! Left lots of small nitpicks to keep our codebase beautiful and shiny forever 🤩

src/lib/utils.ts
trevorkw7: I took a deeper dive into this and found that with SSR, this might cause issues since window isn't defined on server side. Although the way this hook is written solves this window undefined issue by waiting until after the component is mounted using `useEffect`,  which happens after hydration, this also led me to find another problem with depending on hook like this for rendering the navbar that also occurs on the main website. 

SSR serves a temporary fully-formated HTML document before the running client-side JS and comparing these two through hydration. I think we're serving the desktop navbar in the initial HTML document until this hook mounts, after which we can check window size and then hydrate with the mobile navbar. This mismatch causes the bar to flash from desktop to mobile on page load. 

What are your thoughts on this? Should we try using media queries instead to see if it would solve this SSR hydration mismatch?

farisashai:
I like the cool hover effect! 

2 ideas to consider to tweak it a little bit:
- What do you think about hugging the width on the `mobileNav` div and filling the width on each nav item so that we can ensure all of the hover background shapes have the same length?

Screenshot 2023-05-29 at 9 41 58 PM
Screenshot 2023-05-29 at 9 43 16 PM

- Maybe a gap between each nav items so that the hover backgrounds aren't touching each other?
Screenshot 2023-05-29 at 9 45 07 PM

farisashai:
Maybe some input from other people on including the bottom navbar padding under the toggle? I think the Figma mockup didn't have any padding there.

Screenshot 2023-05-29 at 9 47 57 PM

raymosun (changes requested):
Navbar is looking great so far! I think the `useWindowSize()` hook will be really useful too. 

* One bug I'm encountering is it seems like the dark mode toggle doesn't work on the very first click (this happens when i reopen in incognito with no cookie/localstorage). Also, when the site is opened in dark mode to match the device theme, it looks like the switch isn't updated.

Some other minor points:

* Can you bring back your fantastic sun/moon from the original portal, or maybe label the switch with "Dark mode" or "Theme" in the meantime? It's not super obvious what the switch is supposed to do.

* Let's consider a clickaway feature for the navbar, so that touching the page below the navbar will close it.

* Let's think about making the hitbox for the menu button bigger, so it's easier to tap on mobile. I'm thinking it can probably fill up the whole top left corner.
image

trevorkw7 (commented):
Suggested changes have been all addressed besides making the touch target bigger, which we discussed as unnecessary after extensive mobile browser testing.

raymosun (approved):

src/components/layout/Navbar/index.tsx
```diff
@@ -1,18 +1,50 @@
...
 }
 const Navbar = ({ user }: NavbarProps) => {
+  const size = useWindowSize();
+  const headerRef = useRef<HTMLHeadElement>(null);
+
+  const [menuOpen, setMenuOpen] = useState(false);
+  const toggleMenu = useCallback(() => setMenuOpen(prevState => !prevState), []);
+
+  const handleClickOutside = useCallback((event: MouseEvent) => {
+    if (headerRef.current && !headerRef.current.contains(event.target as Node)) {
```
farisashai, line 27: This is an interesting solution for clicking off the header.

There's another approach I've seen commonly where you have a separate layer at a z-index above the page content and beneath the header with a transparent opacity that fills up the entire screen and add the onClick to that layer instead to toggle off (and hide that layer when the header is closed).

I'm not sure if you're tried any alternatives but before you go with this approach, I just wanted to get your thoughts on which one you prefer and why. Since the header is on all pages, would this current approach interfere with any other event targets bubbling up or anything like that.

src/components/layout/Navbar/index.tsx
```diff
@@ -1,18 +1,50 @@
...
+      setMenuOpen(false);
+    }
+  }, []);
```
farisashai, line 30: Should this have the headerRef as a dependency (just for initialization at least)?

src/components/layout/Navbar/index.tsx
```diff
@@ -1,18 +1,50 @@
...
+  useEffect(() => {
+    document.addEventListener('click', handleClickOutside);
+
+    return () => {
+      document.removeEventListener('click', handleClickOutside);
+    };
+  }, [handleClickOutside]);
+
+  // Switch to mobile less than breakpointMd
+  const isMobile = (size.width ?? 0) <= config.cssVars.breakpointMd;
```
farisashai, line 41: Are there any legitimate cases where the size could be undefined or is this just TypeScript complaining?

src/components/layout/Navbar/index.tsx
```diff
@@ -1,18 +1,50 @@
...
+
+  // If they go back to desktop size, don't keep the menu open
+  useEffect(() => {
+    if (!isMobile) setMenuOpen(false);
```
farisashai, line 45: This is a really simple check, I'm wondering if we need it or if we can just use the `isMobile` variable directly to impact whether the header is open or not? See comment on line 92 in this file

src/components/layout/Navbar/index.tsx
```diff
@@ -50,6 +88,40 @@ const Navbar = ({ user }: NavbarProps) => {
           </Link>
         </nav>
       </div>
+      {/* Mobile Menu Dropdown */}
+      <div className={styles.mobileNav} data-open={menuOpen}>
```
farisashai, line 92: If we just modified this to include the mobile requirement, like `data-open={isMobile && menuOpen}`, would that resolve with less code and remove the useEffect above or am I missing something else

farisashai (approved):
Functionality and HTML/CSS structuring looks solid - I just left a few questions for you if you think they're worth addressing, otherwise good to merge whenever you want. 

(I'm not going to read your or Sean's CSS heavily anymore I will just trust yall to know what you're doing otherwise you have to deal with it in the future)

PR #30, opened 2023-05-08
Initial profile page design

### LINKED ISSUE #29: Create User Profile Page
Implement public profile page at `/u/[handle]`

raymosun:
# Info

Implements #29. (If there is no issue for this pull request yet, please create one or
delete this line if the pull request is for a very minor tweak).

# Description

motivation or context.

Added layout for profile page based on preliminary designs.

## Changes

Modifies profile page to begin implementing information display. Adds styles for `u/[handle]` page.

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [ ] locally on Desktop.

# Checklist

- [ ] I have documented my code's `src/lib` functions and commented hard to understand areas
      anywhere else.

# Screenshots

Currently looks like this 
image


src/pages/u/[handle].tsx
```diff
@@ -4,14 +4,51 @@ import withAccessType from '@/lib/hoc/withAccessType';
...
+import Image from 'next/image';
 import type { GetServerSideProps } from 'next/types';
 
 interface UserProfilePageProps {
   user: PublicProfile;
 }
 
 const UserProfilePage = ({ user }: UserProfilePageProps) => {
-  return <pre>{JSON.stringify(user, null, 2)}</pre>;
+  // return <pre>{JSON.stringify(user, null, 2)}</pre>;
```
Comment (farisashai) [inline src/pages/u/[handle].tsx]: you can delete this fully now that we're actually making it

src/pages/u/[handle].tsx
```diff
@@ -4,14 +4,51 @@ import withAccessType from '@/lib/hoc/withAccessType';
...
+  return (
+    <div className={styles.profilePage}>
+      <div className={styles.cardWrapper}>
+        <div className={styles.banner} />
+        <div className={styles.profileCard}>
+          <div className={styles.cardHalf}>
+            <div className={styles.horizontal}>
+              <Image
+                src={user.profilePicture ?? defaultProfilePicture}
```
Comment (farisashai) [inline src/pages/u/[handle].tsx]: Something to think about if you want to include it in this PR with a utility function: How can you randomly provide an image for all ACM users that don't have one uploaded out of the full set of 9 (i think) default images we have

src/styles/pages/u.module.scss
```diff
@@ -0,0 +1,81 @@
+@use 'src/styles/vars.scss' as vars;
+
+.profilePage {
+  margin: auto;
+  max-width: 960px;
+  position: relative;
+  width: 70%;
```
farisashai: I think we might be able to do
```
width: 100%;
max-width: 768px; (use the actual variable for breakpointMd)
```
to achieve a similar effect on desktop with the column width, but easily transition to a full page column view on mobile. let me know what you think about this ^

src/pages/u/[handle].tsx
Comment (farisashai) [inline src/pages/u/[handle].tsx]: Also, doesn't look like this is working with an uploaded pfp for some reason. maybe use the acm testing account as an example with a pfp and your personal acct for an example with no pfp

Screenshot 2023-05-08 at 10 51 02 AM

src/pages/u/[handle].tsx
```diff
@@ -4,14 +4,51 @@ import withAccessType from '@/lib/hoc/withAccessType';
...
+                alt="Profile Picture"
+                className={styles.profilePic}
+              />
+              <div style={{ marginLeft: '1rem' }}>
+                <h1>{`${user.firstName} ${user.lastName}`}</h1>
+                <h3>@{user.handle}</h3>
+                <h2>Rank Here</h2>
+              </div>
+            </div>
+            <div className={styles.iconBox}>icon icon icon icon</div>
```
Comment (farisashai) [inline src/pages/u/[handle].tsx]: message me if you need help getting data to test the social media links with. i know we haven't implemented a way to add a link to your profile yet

src/pages/u/[handle].tsx
```diff
@@ -4,14 +4,51 @@ import withAccessType from '@/lib/hoc/withAccessType';
...
+            <h2>Points</h2>
+            <h2>Class of {user.graduationYear}</h2>
+            <h2>{user.major}</h2>
+          </div>
+        </div>
+      </div>
+
+      <div className={styles.about}>
+        <h1>About me</h1>
+        <p>{user.bio ?? <i>Nothing here...</i>}</p>
```
Comment (farisashai) [inline src/pages/u/[handle].tsx]: The `??` operator is for nullish coalescing and it's a good idea when we're trying to check if the value is completely missing (null or undefined) and use a default value but in this case, we actually don't want that because it may be an empty string, which wouldn't correctly show the default.

Screenshot 2023-05-08 at 10 56 29 AM

farisashai (dismissed):
i know this is nowhere near done but left a few comments - good progress so far 👍

src/styles/pages/u.module.scss
farisashai, line 1: Let's rename this file to `user-profile.module.scss` instead. The other file has to be named specifically to define the route but we should keep this file name readable for anyone looking at it.

farisashai:
Whoops I don't know why I clicked approve

src/pages/u/[handle].tsx
Comment (raymosun) [reply src/pages/u/[handle].tsx]: fixed in Clean up code 1

src/styles/pages/u.module.scss
raymosun: fixed in Clean up code 1

src/pages/u/[handle].tsx
Comment (raymosun) [reply src/pages/u/[handle].tsx]: fixed in Clean up code 1

src/pages/u/[handle].tsx
Comment (raymosun) [reply src/pages/u/[handle].tsx]: fixed in Clean up code 1

src/styles/pages/u.module.scss.d.ts
farisashai, line 1: delete this file since we renamed it

src/styles/pages/user-profile.module.scss
```diff
@@ -0,0 +1,81 @@
...
+        display: flex;
+        justify-content: center;
+      }
+    }
+  }
+  
+  .about {
+    border: 2px solid var(--theme-primary-1);
+    border-radius: 0.5rem;
+    margin-top: 1rem;
```
farisashai: When we design frontend layouts, we have components and containers. 

On the scope of the full page, we can think of each piece on this page, like the `.cardWrapper` and `.about` tags as the components. The tag that contains both of them (`.profilePage`) is the container. 

A component should define how everything inside of itself looks and behaves and allow containers to define how the layout looks when we combine multiple components. Otherwise, if we were to use this component in a different location, it would have an extra margin that isn't relevant to the new page layout.

This basically means, we should try to avoid forcing spacing on individual elements with margin typically and allow the container to handle all layout CSS.

The way that would look on this page is turning `.profilePage` into a column flexbox with a gap of 1rem between elements.

```CSS
.profilePage {
  display: flex;
  flex-direction: column;
  gap: 1rem;
}
```
This gives us the exact same layout but we could theoretically replace the about tag with another similar element and it would fill the space the same way.

src/styles/pages/user-profile.module.scss
```diff
@@ -0,0 +1,81 @@
...
+      top: 0;
+      width: 100%;
+    }
+    
+    .profileCard {
+      background-color: var(--theme-background);
+      border-radius: 0.25rem;
+      box-shadow: 0 2px 4px rgba(0, 0, 0, 0.25);
+      display: flex;
+      margin: 0 2rem;
```
farisashai, line 30: If we hardcode the side margins like this, the content will get extremely squished on mobile views. We can replace this with equal side margins using `margin: 0 auto` and set the width explicitly as `width: 80%; max-width: some rem value` for example or whatever works

farisashai (changes requested):
a few more comments

src/lib/utils.ts
```diff
@@ -27,3 +27,15 @@ export const getMessagesFromError = (errBody: CustomErrorBody): string[] => {
 
   return errBody.errors.map(err => getAllErrMessages(err)).flat();
 };
+
+/**
+ * Get a random default avatar
+ * @returns string path to an avatar .png, suitable for use as src for a Next Image component
+ */
+export const getDefaultAvatarSrc = (): string => {
```
farisashai: If you've seen the current live portal, one funny thing that happens is when you refresh the leaderboard or any other pages, the default user images are constantly re-generated since they're completely random so they switch back and forth for the same users. 

This isn't a big deal, but I was thinking that one way we could address this is by taking in an optional parameter of the user's name or uuid and hashing it to an index mod 9 for one of the profile images so that every user always shows up with the same default image and we still have a good variety with all of the different image options available. Let me know what you think abt this. Maybe we can try it out in a future issue once we have testing data on the leaderboard to see multiple users at once.

farisashai (changes requested):
Can you add a button that links to the edit profile page if the user is viewing their own profile?

raymosun:
This branch won't deploy, possibly because it's very old. The branch is dead 🪦, long live the branch #117.

Issue #52, opened 2023-07-27
Create Storefront Page


raymosun:
Create the storefront at `/store` based on these [figma designs](https://www.figma.com/file/cgbeW6vrqgxW0i1vpbUn1C/Store-(Redesign-2023)?type=design&node-id=654-2155&mode=design&t=jCa9Nm6GIE583hHg-0) (check the 'ready for dev' sections).

Requires #54 and #53

This issue encompasses
- Desktop layout for storefront
- Mobile layout for storefront
- Help Modal for first time users
- Toggle between 'Collections' and 'All Items' viewing modes
  - Collection Card component (Browse collections)
  - Collection Slider component (Browse all items)

![image](https://github.com/acmucsd/membership-portal-ui-v2/assets/30709485/db0a28e8-df25-411c-b4a5-fea1346e1ab3)

SheepTester:
https://github.com/acmucsd/membership-portal-ui-v2/assets/22133785/52758e12-f701-4dd5-b0ce-2bad82046589

I implemented the mobile help modal pagination thing with pure CSS using scroll snapping, which feels kind of sussy so hopefully it works in not-Chrome

farisashai:
Closed in #64

PR #59, opened 2023-08-02
Store item card and navbar components

### LINKED ISSUE #53: Create Store Item Card
Implement Store Item Card component based on these [figma designs](https://www.figma.com/file/cgbeW6vrqgxW0i1vpbUn1C/Store-(Redesign-2023)?type=design&node-id=654-2155&mode=design&t=jCa9Nm6GIE583hHg-0) (check the 'ready for dev' sections).

image

### LINKED ISSUE #54: Implement Store Navbar
Implement the store-specific navbar component used on all store pages based on these [figma designs](https://www.figma.com/file/cgbeW6vrqgxW0i1vpbUn1C/Store-(Redesign-2023)?type=design&node-id=654-2155&mode=design&t=jCa9Nm6GIE583hHg-0) (check the 'ready for dev' sections). 

image

SheepTester:
# Info

Closes #53 and closes #54, which are blockers for most of the store page issues.

# Description

motivation or context.

I added a store item card component and a store navbar component.

## Changes

- Added an `ItemCard` component which takes an `image` URL, `title`, and `cost`. Presumably, these cards will link you to the item page, so there's also an optional `href` prop. I noticed that the designs had a hidden "Out of stock" message, so I also added an optional `inStock` prop which defaults to true that, when set to false, will show "Out of stock" next to the price.

  ```html
  <ItemCard
    image="https://acmucsd.s3.us-west-1.amazonaws.com/portal/profiles/07789a16-8326-4edc-ad2d-fc6193cd1ee3.jpg"
    cost={42069}
    title="Nishant (not for sale)"
    href="/leaderboard"
    inStock={false}
  />
  ```
  - The Figma uses `rgba(228, 228, 228, 0.5)` as the card's shadow color, but this looks like a very unpleasant light grey shadow in dark mode. However, in themes.scss, `--theme-accent-line-2-transparent` is still the same light grey shadow. I could instead use `--theme-accent-line-1-transparent`, which looks fine in dark mode, but in light mode the shadow is very jarring. I added a CSS variable `--theme-shadow` that uses a light grey shadow in light mode and the black shadow in dark mode.
- Added a `Navbar` component which takes the user's `balance` and optionally, `backUrl`. If there is no page to go back to, then leaving `backUrl` unspecified will make the back button go away.

  Ideally, `Navbar` would be able to get the user's balance by itself, but I don't know if it can access the `user: PrivateProfile` on its own.

  ```html
  <Navbar balance={user.credits} backUrl="/store" />
  ```
- I also added a helper component `Diamonds` which renders the number of diamonds with a comma (using the [Internationalization API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat)) and includes the little blue diamond. The diamond icon has a hidden "diamond" text inside it so screenreaders can read it out loud.

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots

Please include a screenshot of your Cypress testing suite passing successfully. no

I put the components on the storefront page just to demo them; the storefront page isn't done yet

Dark mode, desktop:

Dark mode, desktop

Light mode, mobile:

Light mode, mobile


src/components/store/Diamonds/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+interface DiamondsProps {
+  count: number;
+  className?: string;
+}
+
+const Diamonds = ({ count, className }: DiamondsProps) => {
+  return (
+    <span className={className}>
+      {numberFormat.format(count)}
+      <span className={styles.diamond}>diamonds</span>
```
raymosun: image

Nit: the invisible "diamonds" text becomes visible when highlighted! Not a huge deal, but you can fix it by targeting `.diamond::selection` if you want.

src/components/store/Navbar/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+const Navbar = ({ balance, backUrl }: NavbarProps) => {
+  return (
+    <nav className={styles.navbar}>
+      {backUrl && (
+        <Link href={backUrl} className={styles.back}>
+          <svg width="12" height="18" viewBox="0 0 12 18" aria-hidden="true">
+            <path d="M10.1911 18L0 9L10.1911 0L12 1.5975L3.61783 9L12 16.4025L10.1911 18Z" />
+          </svg>
```
raymosun: Can you move this svg to its own file in `/public/assets` so it's more reusable

src/components/store/Navbar/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+        </Link>
+      )}
+      <div className={styles.rightSide}>
+        <span>
+          <strong>Balance:</strong> <Diamonds count={balance} />
+        </span>
+        <Link href="/store/cart" className={styles.navlink}>
+          Cart
+        </Link>
+        <Link href="/store/orders" className={styles.navlink}>
```
raymosun: Can you make variables for these links in `/src/lib/config.ts` and import those so it's easier to update the routes if we ever need to.

image

raymosun (changes requested):
Hey, nice job on this. I really like the implementation of `Diamonds` and the props setup for the other components 🤑 

 - I think it's fine to pass in the credits from the page. Accessing the private profile again in the navbar would probably require an additional API call.

 - One other thing is can you look into modifying the history with the back button instead of just using a link. Currently, clicking the back button will push another page to your history so pushing the custom back button and then pushing the browser back button takes you "forward" to the page you were on. It might be okay to just use `history.back()` have the back button mirror the browser back button, with maybe some consideration for not "backing" out of the store.

SheepTester:
I added an optional `description` prop to `ItemCard` because I think the store item card and [storefront collection card](https://github.com/acmucsd/membership-portal-ui-v2/issues/52) look very similar, and I don't want to have repetitive code.

I also addressed @raymosun's feedback:
1. The back button uses `history.back()` now (well, the one provided by `next/router`).
   
   Currently, it's up to each page whether to show the back button. I wanted to try reading history from `next/router`, but it doesn't look like they provide it. Otherwise, I'd only show the back button if the previous page is under `/store` but isn't the storefront page.
   
   If someone linked directly to an item page, the back button would take the visitor out of the website rather than to the storefront page, which might be a bit jarring, but I'm not sure what user expectations are for a back button when you're linked directly to a sub page.
2. I just pushed "diamonds" out of the square with a bit of padding rather than changing the highlight color.
3. The back icon is in asserts/icons/back-chevron.svg now. I was considering replacing back-arrow.svg, which is larger and has a shaft and isn't used anywhere, but I decided to leave it be.
4. The cart links were moved to config.ts.

raymosun (approved):

raymosun:
Hey Sean, thanks for addressing the changes! I think prepping for the component to handle item card and collection card is a good idea.

So, the only thing that's still iffy is the back button, but I think it's good enough for now so we can go ahead and merge so the other store pages aren't blocked and maybe address it with a later issue.

SheepTester:
image
bruh

src/components/store/ItemCard/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+    <Card href={href ?? ''} className={styles.itemCard}>
+      <div className={styles.imageWrapper}>
+        <Image src={image} alt={title} fill />
+      </div>
+      <div className={styles.details}>
+        <p className={styles.title}>{title}</p>
+        {description && <p>{description}</p>}
+        {(cost || !inStock) && (
+          <p className={styles.cost}>
+            {cost && <Diamonds count={cost} />}{' '}
```
farisashai:
```suggestion
            {cost && <Diamonds count={cost} />}&nbsp;
```

nit: I like to avoid the weird `{' '}` whitespace thing by adding the HTML escape character for a non-breaking space manually

src/components/store/Navbar/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+  showBack?: boolean;
+}
+
+const Navbar = ({ balance, showBack }: NavbarProps) => {
+  const router = useRouter();
+
+  return (
+    <nav className={styles.navbar}>
+      {showBack && (
+        <button type="button" className={styles.back} onClick={() => router.back()}>
```
farisashai: A page router history approach might work really work for some situations but it might actually be a little funky here.

Example: I open a new tab and enter the URL to an item page directly that contains the back button. If i'm correct, the back button wouldn't do anything in this case when we really want it to take us to the store home page. 

Can you handle this differently somehow?

public/assets/icons/back-chevron.svg
farisashai, line 1: Uhh I know the technical name is actually chevron but can we simplify it for readability to arrow or something bc I don't think anyone knows what that means.

src/components/store/Diamonds/index.tsx
```diff
@@ -0,0 +1,18 @@
+import styles from './style.module.scss';
+
+const numberFormat = new Intl.NumberFormat('en-US');
+
+interface DiamondsProps {
+  count: number;
+  className?: string;
+}
+
+const Diamonds = ({ count, className }: DiamondsProps) => {
```
farisashai:
```suggestion
const Diamonds = ({ count, className }: DiamondsProps) => <span></span>
```


optional nitpick: immediate function return values can be inlined (even for components).

with proper indentation though, not all on one line

src/components/store/ItemCard/style.module.scss
```diff
@@ -0,0 +1,43 @@
+@use 'src/styles/vars.scss' as vars;
+
+.itemCard {
+  background-color: var(--theme-background);
+  border-radius: 0.625rem;
```
farisashai: nit: we usually just use `px` for values this precise below 1rem

src/components/store/Navbar/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+import Link from 'next/link';
+import { useRouter } from 'next/router';
+import styles from './style.module.scss';
+
+interface NavbarProps {
+  balance: number;
+  showBack?: boolean;
+}
+
+const Navbar = ({ balance, showBack }: NavbarProps) => {
```
farisashai: Can we rename this component differently to StoreHeader or StoreNavbar or something to reduce ambiguity since the main Navbar component has the same name

src/components/store/Navbar/style.module.scss
```diff
@@ -0,0 +1,41 @@
...
+  .back {
+    align-items: center;
+    background: none;
+    color: var(--theme-text-on-background-1);
+    display: flex;
+    font: inherit;
+    font-weight: bold;
+    gap: 8px;
+
+    @media (max-width: vars.$breakpoint-sm) {
```
farisashai:
```suggestion
    @media screen and (max-width: vars.$breakpoint-sm) {
```

nit but good practice to follow for media queries. 
`@ media print` would be css that applies for printing a webpage for example

src/components/store/index.ts
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable import/prefer-default-export */
```
farisashai: does it still throw this error when you have multiple exports? if not, you can remove it

src/pages/store/index.tsx
```diff
@@ -1,9 +1,45 @@
...
+          gap: '10px',
+          gridTemplateColumns: 'repeat(auto-fill, minmax(300px, 1fr))',
+          marginTop: '32px',
+        }}
+      >
+        <ItemCard
+          image="https://acmucsd.s3.us-west-1.amazonaws.com/portal/profiles/07789a16-8326-4edc-ad2d-fc6193cd1ee3.jpg"
+          cost={42069}
+          title="Nishant (not for sale)"
+          href="/leaderboard"
```
farisashai:
```suggestion
          href="/store/items/uuid"
```
Can you mock more realistic data even if it doesn't exist and is the same for all items

src/pages/store/index.tsx
```diff
@@ -1,9 +1,45 @@
...
+          inStock={false}
+        />
+        <ItemCard
+          image="https://acmucsd.s3.us-west-1.amazonaws.com/portal/profiles/a21d9a9d-0da5-479d-9183-9781be4f9daf.jpeg"
+          cost={10}
+          title="Critter"
+        />
+        <ItemCard
+          image="https://acmucsd.s3.us-west-1.amazonaws.com/portal/profiles/cf6ca1b7-cd3b-4da5-8d2b-28ea55084a06.png"
+          title="The Crewmate Collection"
```
farisashai: Since you mentioned to want to demo the collection images, can you make a temporary section for those differently than the demo store items to reduce some confusion

farisashai (changes requested):
Components look good so far! Some questions and comments and a few nits to address.

farisashai:
Can you include edge case screenshots of handling an item card with a description too long?

farisashai:
image

fitting content is a little funky on iPhone SE size, up to you if you want to address it. nobody is going to have 10 million points anytime soon 😮

src/components/store/Navbar/index.tsx
SheepTester: My original approach was that each page would provide the URL to return to as a prop, so a collection page would pass the storefront page, an item page could pass a collection URL/the storefront page, etc.

Raymond suggested using `history.back()`. Ideally, I'd hide the button if the previous page isn't inside `/store`, e.g. the user came from another part of the ACM website or outside.

A partial solution is using `document.referrer`, but NextJS's router to some degree acts like a SPA, so the page never reloads, and the value of `document.referrer` stays the same while the user navigates between pages on the website.

NextJS router [doesn't provide components direct access to the history](https://github.com/vercel/next.js/discussions/36723) of routes within the app, and tracking paths seems involved and potentially buggy since you'd also have to track when the user uses the browser's forward and back buttons.

My personal opinion is that the back button doesn't add any functionality the user doesn't already have (the browser already has a back button), and it adds confusion because the mental model for a back button inside an app vs. a website is very different (I'd expect clicking on a store icon then going back goes to the previous page on a website, and the home page in an app). I think that we should remove the back button altogether, or replace it with breadcrumbs `Store > Collection name > Item name`.

SheepTester:
> Can you include edge case screenshots of handling an item card with a description too long?

image

When one card has an insanely long description, the text wraps and the other cards' height match the height of the longest card. I believe this is consistent with the Figma (below), and personally I think it's a bit better than having staggered card heights. However, if having a lot of blank space on shorter cards isn't desirable, then changing it is a single-line change (`align-items: flex-start`)

image

farisashai (approved):
the store nav back button needs to function more robustly in any possible scenario and the cards shouldn't break the height that much with long text content but i'll approve and leave those to you to fix later

PR #77, opened 2023-10-14
Leaderboard sorting and pagination 

SheepTester:
# Info

Closes #63.

# Description

What changes did you make? List all distinct problems that this PR addresses. Explain any relevant
motivation or context.

I added a fancier dropdown and pagination.

I copied the pagination code from the ACM dev team take-home btw 

## Changes

- added custom dropdown
  - it has past week etc options
- it filters by year
  - unlike the current website, filtering by academic year is a bit more dumb in this implementation. The current(old) website imports a npm package maintained by ACM that hard codes the quarter start and end dates to determine the time range for a school year. Here, I just assumed the school years change on August 1
- I copied my pagination code from the take-home assignment so loading the all time leaderboard no longer crashes my browser
- the my position button disables itself when you are not on the leaderboard

# Type of Change

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as
      expected)
- [ ] Logistics Change (A change to a README, description, or dev workflow setup like
      linting/formatting)
- [ ] Continuous Integration Change (Related to deployment steps or continuous integration
      workflows)
- [ ] Other: (Fill In) 

# Testing

I have tested that my changes fully resolve the linked issue ...

- [x] locally on Desktop.
- [ ] locally on mobile - use https://ngrok.io to get a copy on a mobile device
- [ ] on the live deployment preview on Desktop.
- [x] on the live deployment preview on Mobile.
- [x] I have run and passed all new and existing Cypress tests. Add screenshots below.

# Checklist

- [x] I have performed a self-review of my own code.
- [x] I have followed the style guidelines of this project.
- [x] I have documented my code's `src/lib` functions and commented hard to understand areas
      anywhere else.
- [x] My changes produce no new warnings.

# Screenshots

Please include a screenshot of your Cypress testing suite passing successfully.

If you made any visual changes to the website, please include relevant screenshots below.

Screenshot_20231014-155355

Issue #63, opened 2023-08-12
Leaderboard sorting


raymosun:
Add sorting with dropdown for leaderboard page.

![image](https://github.com/acmucsd/membership-portal-ui-v2/assets/30709485/fc088df8-f28a-4a4a-9dc2-997266b6cebd)

Tentatively, have the options like this. Can also discuss other optinos.
![image](https://github.com/acmucsd/membership-portal-ui-v2/assets/30709485/d55bab87-fff5-49c8-b03c-0e77734935dc)

PR #83, opened 2023-10-21
Revamp admin event UI, add searching and community filtering

alexzhang1618:
# Info

Closes **[ISSUE NUMBER]**. (If there is no issue for this pull request yet, please create one or
delete this line if the pull request is for a very minor tweak).

# Description
- Change community selection into a common enum type
- Revamp manage events screen with both row and card views
- Add searching and community filtering for manage events screen
- Add date parsing utils for event display
- Display cover image on edit event page
- Disable Notion dropdown while options are loading
- Automatically fill fields when an autofill option is selected
- Add event pagination
- Add max-width
- TODOS:
  - Explore editing the cover image of an event
  - Explore event cover image cropping on upload
  - Implement generate ACMURL (when added to Klefki)

## Changes

- [Fill in here]

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

- [ ] I have documented my code's `src/lib` functions and commented hard to understand areas
      anywhere else.

# Screenshots


src/components/admin/event/EventDetailsForm/index.tsx
```diff
@@ -287,6 +286,14 @@ const EventDetailsForm = (props: IProps) => {
...
           </>
         )}
+        {initialValues.cover ? (
+          <>
+            <span>Cover Image</span>
+            <div className={style.coverImageContainer}>
+              <Image src={initialValues.cover || ''} alt="Event cover" objectFit="cover" fill />
+            </div>
+          </>
+        ) : null}
```
SheepTester, line 296: It would be nice if it showed the selected cover image as a preview on the create event page, and it let you change the cover image on the edit event page.

Alternatively, perhaps it could show a preview of the entire event card with the title and cover image etc under the form

src/components/admin/event/EventDetailsForm/index.tsx
```diff
@@ -287,6 +286,14 @@ const EventDetailsForm = (props: IProps) => {
...
       </div>
 
       <div className={style.submitButtons}>
```
SheepTester, line 299: I'm not sure if a clear form button is a good idea since it would be quite frustrating to accidentally click it instead of "create event," and it's easy enough just to reload the page to reset it.

[MDN doesn't recommend having a reset button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/reset)

> **Note**: You should usually avoid including reset buttons in your forms. They're rarely useful, and are instead more likely to frustrate users who click them by mistake (often while trying to click the [submit button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit)).

I can imagine the discard changes button being useful, but it should warn the user (eg with a `confirm` dialog) before resetting everything. Likewise for delete event and other destructive buttons

src/lib/utils.ts
```diff
@@ -79,3 +79,31 @@ export const getUserRank = (user: PublicProfile): string => {
...
+ * Given ISO-8601 strings for the start and end date of a location,
+ * this function returns a formatted date for us to display on each event.
+ * Example outputs:
+ *  - Apr 25
+ *  - Jun 11
+ */
+export const formatDate = (date: string): string => {
+  const startDate = new Date(date);
+  const month = MONTHS[startDate.getMonth()];
+  return `${month} ${startDate.getUTCDate()}`;
```
SheepTester: You probably should use `.getUTCMonth` in tandem with `getUTCDate`.

May I suggest using the Intl API for this?

```js
> new Date('2023-01-02').toLocaleString(['en-US'], { month: 'short', day: 'numeric', timeZone: 'UTC' })
'Jan 2'
```

SheepTester: The current implementation is prone to time zone issues, actually:

image

The hackathon event shows as "Oct 1," which in fact it's on Oct 31. It's just that Oct 31 6 pm here is Nov 1 in England.

JavaScript's ISO date parsing is a bit broken, so just saying "ISO-8601 string" in the JSDoc isn't enough. If it's just the date and no time (eg `2023-10-21`), JavaScript treats the date as at 12 am UTC. Otherwise, it uses local time.

```js
> new Date('2023-10-21T18:46')
Sat Oct 21 2023 18:46:00 GMT-0700 (Pacific Daylight Time)
> new Date('2023-10-21')
Fri Oct 20 2023 17:00:00 GMT-0700 (Pacific Daylight Time)
```

I believe that the event start and end times are taken directly from `<input type="datetime-local">`, which formats dates in local time as `YYYY-MM-DDThh:mm`. So using `getUTCDate` here is wrong.

```suggestion
  const startDate = new Date(date);
  const month = MONTHS[startDate.getMonth()];
  return `${month} ${startDate.getDate()}`;
```

or

```suggestion
  return new Date(date).toLocaleString('en-US', { month: 'short', day: 'numeric' });
```

The website will display the same time to everyone, regardless of their time zone. But this shouldn't be a problem post-Covid since everyone should be in San Diego now

src/lib/utils.ts
```diff
@@ -79,3 +79,31 @@ export const getUserRank = (user: PublicProfile): string => {
...
+    hour: 'numeric',
+    minute: '2-digit',
+  });
+  const endTime = new Date(end).toLocaleTimeString('en-US', { hour: 'numeric', minute: '2-digit' });
+  return `${startTime} - ${endTime}`;
+};
+
+export const formatEventDate = (start: string, end: string): string => {
+  return `${formatDate(start)}, ${formatTime(start, end)}`;
+};
```
SheepTester: If the other functions aren't being used anywhere, then you could just use [`formatRange`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatRange):

```suggestion
const rangeFormat = new Intl.DateTimeFormat('en-US', { month: 'short', day: 'numeric', hour: 'numeric', minute: '2-digit' });
export const formatEventDate = (start: string, end: string): string => {
  return rangeFormat.formatRange(new Date(start), new Date(end));
};
```

This is what its output looks like; it's smart and will only repeat the date or am/pm if it's different:

```js
> new Intl.DateTimeFormat('en-US', { month: 'short', day: 'numeric', hour: 'numeric', minute: '2-digit' }).formatRange(new Date(2021, 9, 21, 18, 0), new Date(2021, 9, 21, 18, 30))
'Oct 21, 6:00 – 6:30 PM'
```

src/pages/admin/event/index.tsx
```diff
@@ -5,21 +5,45 @@ import { EventAPI } from '@/lib/api';
...
 const ManageAllEventsPage: NextPage<IProps> = (props: IProps) => {
+  const router = useRouter();
   const { allEvents, pastEvents, futureEvents } = props;
 
   const [mode, setMode] = useState<'all' | 'past' | 'future'>('future');
+  const [query, setQuery] = useState<string>('');
+  const [viewMode, setViewMode] = useState<'card' | 'row'>('row');
+  const [communityFilter, setCommunityFilter] = useState<string>(NO_FILTER_OPTION);
+  const [currentPage, setCurrentPage] = useState<number>(0);
+  const [paginationSize, setPaginationSize] = useState<'All' | number>('All');
```
SheepTester: Should it default to a smaller number than just all? If there's a lot of events, it probably wouldn't be great if the user had to wait for all the events to render before being able to select a smaller page size.

src/pages/admin/event/index.tsx
```diff
@@ -34,11 +58,83 @@ const ManageAllEventsPage: NextPage<IProps> = (props: IProps) => {
...
+    if (e.target.value === 'All') {
+      setPaginationSize(e.target.value);
+    } else {
+      setPaginationSize(parseInt(e.target.value, 10));
+    }
+  };
+
+  useEffect(() => {
+    setCurrentPage(0);
+  }, [pageSize]);
```
SheepTester: I think the page number should also reset if the user changes the filter or query

src/pages/admin/event/index.tsx
SheepTester: The leaderboard PR #63 also adds pagination, searching, and filters. Perhaps some day we might want to merge them into a common component

src/pages/admin/event/index.tsx
```diff
@@ -34,11 +58,83 @@ const ManageAllEventsPage: NextPage<IProps> = (props: IProps) => {
...
-        <div className={style.buttons}>
+        <h1>Manage Events</h1>
+        <div className={style.viewControls}>
+          <IconButton onClick={() => setViewMode('card')}>
+            <BsFillGrid3X3GapFill className={viewMode === 'card' ? style.active : style.inactive} />
+          </IconButton>
+          <IconButton onClick={() => setViewMode('row')}>
+            <FaThList className={viewMode === 'row' ? style.active : style.inactive} />
+          </IconButton>
+        </div>
```
SheepTester, line 89: It might be nice if it could save the user's preferred view mode in localStorage or in the URL so that it doesn't reset to list view when they edit an event and go back to the list

src/styles/pages/manage-events.module.scss
```diff
@@ -1,19 +1,105 @@
...
+  }
+
+  .controls {
+    display: flex;
+    justify-content: space-between;
+
+    .controlsLeft {
+      display: flex;
+      flex-direction: row;
+      gap: 1rem;
```
SheepTester: It doesn't look so great on mobile. I feel like there might be some events team people who'll be editing their event while on the go, so I think it's important that it's at least usable on mobile. Adding `flex-wrap: wrap;` to `.eventInfo`, `.controls` (as well as `gap: 1rem`), and its two children should help:

Current | With `flex-wrap`
--- | ---
image | image

SheepTester (changes requested):
Perhaps a gap in the row view might make it look nice, or alternating background colors like with the leaderboard?

Current | With gap | No border, alternating rows
--- | --- | ---
image | image | image

I experienced an issue trying to create an event where if my cover photo was too large, it'd say that it's too large, but then after uploading a small file, it gives an error message about a duplicate key for the check-in code. It created the event with a broken image URL for the cover image even though it was too large, and I don't think there's a way to change the cover photo afterwards other than deleting and recreating it

Also, formatting dates is a bit broken (see below); the simplest fix is just to use `getDate` instead of `getUTCDate`

🤠

src/pages/admin/event/index.tsx
alexzhang1618: That makes sense, if you want you could refactor #63 to make some of the stuff common components and I can refactor this branch to use that (or I can do that in a follow-up PR)

alexzhang1618:
Changes so far:
- Added flex-wraps to fix mobile views
- Fixed date parsing (I only needed the range times, so I removed some of the extraneous stuff)
- Changed default pageSize to 5
- Removed the 'resetForm' button in edit event
- Changed the backgrounds for row/card views to be more visually appealing

src/components/admin/event/EventDetailsForm/index.tsx
alexzhang1618: Yeah, I think I'll try to address some of the image stuff/editing bugs in a future PR where I'll also implement generateACMURL if that works

alexzhang1618: Additionally, I like the event card preview idea, I can add that when we implement the component down the line

SheepTester:
fyi, since #77 was merged 🦀🦀, we now have some 🅱️ased components like

- [`Dropdown`](https://github.com/acmucsd/membership-portal-ui-v2/tree/main/src/components/common/Dropdown)
- [`PaginationControls`](https://github.com/acmucsd/membership-portal-ui-v2/tree/main/src/components/common/PaginationControls)

which might be helpful for this PR

src/pages/admin/event/index.tsx
```diff
@@ -5,21 +5,45 @@ import { EventAPI } from '@/lib/api';
...
 interface IProps {
   allEvents: PublicEvent[];
   pastEvents: PublicEvent[];
   futureEvents: PublicEvent[];
 }
 
+const NO_FILTER_OPTION = 'All Communities';
+const communityFilterOptions = [
+  NO_FILTER_OPTION,
+  ...(Object.keys(Community) as Array<keyof typeof Community>).map(key => Community[key]),
```
raymosun, line 27: seeking to learn: what's the advantage of this over `...Object.values(Community)` ?

src/pages/admin/event/index.tsx
```diff
@@ -34,11 +58,83 @@ const ManageAllEventsPage: NextPage<IProps> = (props: IProps) => {
...
+  const filteredEvents = visibleEvents
+    .filter(event => communityMatch(event, communityFilter))
+    .filter(event => event.title.toLowerCase().includes(query));
+
+  const pageSize = paginationSize === 'All' ? filteredEvents.length : paginationSize;
+
+  const nextPage = () => {
+    if ((currentPage + 1) * pageSize < filteredEvents.length) {
+      setCurrentPage(current => current + 1);
+    }
```
raymosun: If we're using a function update to avoid using a stale value of `currentPage`, might as well go all the way and check the value in the function too, e.g. `setCurrentPage(current => (current+ 1) * pageSize < filteredEvents.length ? current + 1 : current)` to ensure users can't get to an invalid page by clicking too quickly.

src/pages/admin/event/index.tsx
```diff
@@ -34,11 +58,83 @@ const ManageAllEventsPage: NextPage<IProps> = (props: IProps) => {
...
+  };
+
+  const prevPage = () => {
+    if (currentPage > 0) {
+      setCurrentPage(current => current - 1);
+    }
```
raymosun: Ditto from `nextPage()`, e.g. `setCurrentPage(current => current > 0 ? current - 1 : current)`

src/pages/admin/event/index.tsx
```diff
@@ -34,11 +58,83 @@ const ManageAllEventsPage: NextPage<IProps> = (props: IProps) => {
...
+  };
+
+  const changePageSize = (e: ChangeEvent<HTMLSelectElement>) => {
```
raymosun: Nit: in keeping with naming conventions for event handlers, let's call this `onChangePageSize`

src/pages/admin/event/index.tsx
```diff
@@ -61,12 +157,40 @@ const ManageAllEventsPage: NextPage<IProps> = (props: IProps) => {
...
+          <select
+            id="pageSize"
+            value={pageSize}
+            onChange={e => changePageSize(e)}
+            className={style.paginationSizeSelect}
+          >
+            {['All', 5, 10, 20].map(size => (
+              <option key={size} value={size}>
+                {size}
+              </option>
```
raymosun: These options are currently invisible in dark mode. Let's target `option` elements under `.paginationSizeSelect` to give them a theme color.

image

src/styles/pages/manage-events.module.scss
```diff
@@ -1,21 +1,115 @@
...
+
+        .searchInput {
+          background: #fff;
+          border: 0;
+          color: vars.$light-primary-2;
+        }
+
+        .searchInput::placeholder {
+          color: vars.$light-primary-2;
+        }
```
raymosun: I'm not a fan of how the placeholder has the same color as input text. The input looks identical to the buttons and it doesn't feel like something I can type in.

image

src/components/admin/event/EventDetailsForm/index.tsx
```diff
@@ -287,6 +286,14 @@ const EventDetailsForm = (props: IProps) => {
             </EventDetailsFormItem>
           </>
         )}
+        {initialValues.cover ? (
```
raymosun, line 289: tiny nit: it should suffice to have `intialValues.cover && (...)`

src/pages/_app.tsx
```diff
@@ -30,10 +38,12 @@ export default function MyApp({ Component, pageProps }: AppProps<InitialPropInte
...
-        <ToastContainer />
-        <PageLayout user={pageProps?.user}>
-          <Component {...pageProps} />
-        </PageLayout>
+        <MUIThemeProvider theme={theme}>
+          <ToastContainer />
+          <PageLayout user={pageProps?.user}>
+            <Component {...pageProps} />
+          </PageLayout>
+        </MUIThemeProvider>
```
raymosun, line 46: question: have we been using MUI for portal v2, or does this PR introduce its use?

raymosun (changes requested):
The UI looks fire, super convenient for managing events 🔥. Leaving some comments:

1. Let's bring back the `Autofill Form` button functionality, so data doesn't get overwritten as soon as you change the notion event.

image

2. Let's give the pagination buttons and pagination-size-dropdown pointer cursors to make it clear they're interactable. 

image

3. Let's consider adding a pagination bar to the bottom of the table as well.

trevorkw7 (commented):
Apologies on the delay for this review but great job on the massive facelift on the UI! I the grid and list layout options specifically. I don't have much to say on the UI itself but lets address some nits about the implementation we have here.

1. There seems to an interesting hydration bug that occurs when clicking the view public event page and coming back to the admin page afterwards. It probably has something to do with a discrepancy between the initial server render and some client side updates since we are using SSR but some more investigation needs to be done.
image

2. Similar to what Raymond asked, I was curious why we imported MUI to implement the tooltips and icon buttons along with adding some global styling to the project as a whole? I think Frontend's general sentiment in creating portal v2 was to move away from UI Frameworks like MUI because of the large package sizes they bring and the difficulty there is in customizing them using pure CSS. This was the catalyst towards us building our own component library and the philosophy of only bringing in UI dependencies when building our own is unnecessarily difficult (i.e. calendar pickers) through headless components (react aria, headless ui). In terms of what MUI is being used to implement here, they like things we've already implemented at one point like the Icon Links in our navbar or the tool tips in the store front page.

PR #91, opened 2023-11-14
Edit profile page

### LINKED ISSUE #87: Create Edit User Profile Page
Figma: https://www.figma.com/file/WCc4V3sy05iF55BRum71Mv/Portal-UI-v2-Pages?type=design&node-id=911-5585&mode=design&t=RZ7bjgMRD5oij9PG-4

Design mockup of edit user profile page

See also, #29

SheepTester:
# Info

Closes #87 

# Description

motivation or context.

Created the edit profile page, including its functionality.

I've also added some new components and utility functions, which would be helpful for other parts of the website as well:

- An accessible `Modal` component for general modals.
- A `Cropper` component for cropping user-selected images to a specific aspect ratio. It will also help discourage users from uploading non-images, and it enforces max file sizes by compressing the file as a JPG until it's small enough.
- A custom hook `useObjectUrl` that creates an object URL for the given `File`, and revokes it when it's done.
- An SCSS `accessibly-hidden` mixin for elements that should be tab-accessible but should also pretend to be hidden.

## Changes

- Created edit profile page
- Applied the new utility stuff above to other parts of the site as needed, including adding the cropper to the create event form.

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots

Desktop, dark mode | Mobile, light mode
--- | ---
OVERWRITE ME | OVERWRITE ME

image


src/components/admin/event/EventDetailsForm/index.tsx
```diff
@@ -267,24 +277,48 @@ const EventDetailsForm = (props: IProps) => {
...
+                    setSelectedCover(file);
+                  }
+                }}
               />
             </EventDetailsFormItem>
+            <Cropper
+              file={selectedCover}
+              aspectRatio={1920 / 1080}
+              maxFileHeight={1080}
+              maxSize={config.file.MAX_EVENT_COVER_SIZE_KB * 1024}
```
alexzhang1618, line 301: I like this maxSize thing here as a warning for users!

one thing I think that's worth mentioning is that the max file size can vary between testing.api and prod (for testing it's 256 KB, for prod its 2048 KB), so on testing you might get some weird behavior:
- you try to upload an image that's in between 256 KB and 2048 KB
- the event is created successfully but the image doesn't get uploaded

this isn't a massive issue since it only affects testing and not prod, but it could be worth looking into having this config value switch depending on the deploy/runtime var

src/components/admin/event/EventDetailsForm/index.tsx
```diff
@@ -267,24 +277,48 @@ const EventDetailsForm = (props: IProps) => {
...
-                  },
-                })}
+                accept="image/*"
+                onChange={async e => {
+                  const file = e.currentTarget.files?.[0];
+                  e.currentTarget.value = '';
+                  if (file) {
+                    setSelectedCover(file);
+                  }
+                }}
```
alexzhang1618: Why is this `onChange` async? at a glance the function looks synchronous

src/components/common/Cropper/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+interface CropperProps {
+  file: Blob | null;
+  aspectRatio: number;
+  circle?: boolean;
+  maxFileHeight: number;
+  maxSize?: number;
+  // eslint-disable-next-line no-unused-vars
+  onCrop: (file: Blob) => void;
+  // eslint-disable-next-line no-unused-vars
+  onClose: (reason: 'invalid-image' | 'cannot-compress' | null) => void;
```
alexzhang1618: are these rules still supposed to be disabled here?

src/components/common/Cropper/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+            const blobs = await Promise.all(
+              [1, 0.9, 0.7, 0.4, 0].map(quality => toBlob(canvas, 'image/jpeg', quality))
+            );
+            const firstSmallEnough = blobs.find(blob => blob && blob.size <= maxSize);
+            if (firstSmallEnough) {
+              onCrop(firstSmallEnough);
+            } else {
+              onClose('cannot-compress');
+            }
+          }}
```
alexzhang1618, line 188: 🔥 🔥 🔥 this is epic

src/components/profile/EditField/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+  type?: string;
+  element?: 'input' | 'textarea' | 'select';
+  options?: string[];
+  maxLength?: number;
+  disabled?: boolean;
+  value: string;
+  // eslint-disable-next-line no-unused-vars
+  onChange: (value: string) => void;
+  // eslint-disable-next-line no-unused-vars
+  onBlur?: (value: string) => string;
```
alexzhang1618: same here, can these `eslint-disable`s be removed?

src/components/profile/Preview/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+    case SocialMediaType.LINKEDIN:
+      return <BsLinkedin className={styles.icon} aria-label="LinkedIn" />;
+    case SocialMediaType.PORTFOLIO:
+      return <AiOutlineLink className={styles.icon} aria-label="Portfolio" />;
+    case SocialMediaType.TWITTER:
+      return <BsTwitter className={styles.icon} aria-label="Twitter" />;
+    default:
+      return null;
+  }
+}
```
alexzhang1618: We might want to move this out into a separate component since I think this'll be useful for the actual profile page

src/components/profile/Switch/index.tsx
```diff
@@ -0,0 +1,27 @@
+import { ReactNode } from 'react';
+import styles from './style.module.scss';
+
+interface SwitchProps {
+  checked: boolean;
+  // eslint-disable-next-line no-unused-vars
+  onCheck: (checked: boolean) => void;
```
alexzhang1618: same thing here with eslint-disable

src/lib/api/UserAPI.ts
```diff
@@ -41,3 +55,113 @@ export const getUserByHandle = async (token: string, handle: string): Promise<Pu
...
+  return response.data.user;
+};
+
+/**
+ * Update current user's private profile
+ * @param token Authorization bearer token
+ * @param user Profile changes
+ * @returns User's full profile
+ */
+export const updateCurrentUser = async (
```
alexzhang1618: Thoughts on renaming this function name to `updateCurrentUserProfile` or something? at a glance it looked like this function was switching which user was logged in

src/lib/hoc/withAccessType.ts
```diff
@@ -51,19 +51,21 @@ export default function withAccessType(
...
+    let userAccessLevel: UserAccessType;
 
-    // If no user cookie, try to re-generate one
-    if (!userCookie || !userAccessLevel) {
-      try {
+    try {
+      user = JSON.parse(userCookie);
+      userAccessLevel = user.accessType;
+      // If no user cookie, try to re-generate one
+      if (!userCookie || !userAccessLevel) {
```
alexzhang1618, line 61: seeking to learn: what does this change?

src/pages/profile/edit.tsx
```diff
@@ -0,0 +1,660 @@
...
+              if (
+                firstName !== user.firstName ||
+                lastName !== user.lastName ||
+                handle !== user.handle ||
+                email !== user.email ||
+                major !== user.major ||
+                +graduationYear !== user.graduationYear ||
+                bio !== user.bio ||
+                isAttendancePublic !== user.isAttendancePublic ||
+                newPassword.length > 0
```
alexzhang1618: Can this be bundled with `hasChange`? (i.e.) `hasChange = hasProfileChange && hasSocialMediaChange`

src/pages/profile/edit.tsx
```diff
@@ -0,0 +1,660 @@
...
+          </section>
+          <form
+            className={styles.columnRight}
+            onSubmit={async e => {
+              e.preventDefault();
+
+              let changed = false;
+              if (
+                firstName !== user.firstName ||
+                lastName !== user.lastName ||
```
alexzhang1618: Thoughts on putting this into a `handleSubmit` function to make the JSX section more concise?

src/pages/profile/edit.tsx
```diff
@@ -0,0 +1,660 @@
...
+                          className={styles.fileInput}
+                          type="file"
+                          accept="image/*"
+                          onChange={async e => {
+                            const file = e.currentTarget.files?.[0];
+                            e.currentTarget.value = '';
+                            if (file) {
+                              setPfp(file);
+                            }
+                          }}
```
alexzhang1618: Same question as in EventDetailsForm

alexzhang1618 (commented):
Functionality works great! Thanks for also refactoring the other modals, EventDetailsForm, and the Add to Apple Calendar button 😁 

Some questions/comments I had:

src/components/admin/event/EventDetailsForm/index.tsx
alexzhang1618: just for reference:

on prod: 
- event cover: 2048
- pfp: 256
- banner (don't think we use these currently): 2048
- merchPhoto: 1024
- resume: 2048

on testing/staging:
- event cover: 256
- pfp: 256
- banner (don't think we use these currently): 256
- merchPhoto: 1024
- resume: wasn't specified in caprover for some reason

SheepTester: This is handled by the config file which sets the max size depending on prod vs dev

Regardless, the cropper is guaranteed to close with an image file under the specified file size. If the image is too large, it'll try compressing it with lower and lower quality until it fits under the maximum

src/components/common/Cropper/index.tsx
SheepTester: Yeah, for some reason ESLint thinks the parameters are unused even though they're just part of the type signature. You can see the rule being disabled throughout the codebase

src/components/admin/event/EventDetailsForm/index.tsx
alexzhang1618: oh hm not sure why I was having that bug earlier then. Do we want to not make the updateEvent API call if the image is too big?

src/lib/hoc/withAccessType.ts
SheepTester: This is to prevent locking the user out if their user cookie contains malformed JSON for some reason

I made this change because my initial approach to update the user cookie after changing the profile was to set it to an empty string to force the server to regenerate it

src/components/admin/event/EventDetailsForm/index.tsx
SheepTester: hmm what image were you using?

we could add a check in `AdminEventManager.createNewEvent` to prevent creating an event if the cover is too large, but `Cropper` should make the same check already

alexzhang1618: Does `Cropper` block form submission if the cover is too large? if not, then we should probably enforce a check here before allowing submission

alexzhang1618 (approved):
lgtm 🚢

src/lib/api/UserAPI.ts
```diff
@@ -1,9 +1,23 @@
 import { config } from '@/lib';
+import { UUID } from '@/lib/types';
```
farisashai:
```suggestion
import type { UUID } from '@/lib/types';
```

farisashai (changes requested):
Everything looks great and functional - couple UI nitpicks.  

- [x] I'm on mobile and a dropdown arrows shows up on both the left and right side of the details tag which seems unintended.

- [x] Thoughts on "Edit Profile" -> "Manage Account" since not all of the edits on this page are publicly facing. 

- [x] We have a route on the website `/logout` but no links to it anywhere so far. I think this page makes the most sense to include that link since it wouldn't be used too commonly by most users. 

- [x] I think we have something similar on some other form components but it would help a lot to highlight which fields have been edited and are currently unsaved so I can know before clicking the cancel or save button and making unintended edits.

raymosun (changes requested):
Hey this looks great! The cropper modal is really well done. Just a few concerns: 

1. Is this delete button never available? Maybe we can remove it.
image

2. In dark mode, the placeholder color doesn't have enough contrast. (Also the character count color)
image
image

3. ~Edit: I can't seem to see my current resume on the page~ Editedit: I deleted my cookie and it's fine, dunno what was up with that.

raymosun:
> Everything looks great and functional - couple UI nitpicks.
> 
> 1. I'm on mobile and a dropdown arrows shows up on both the left and right side of the details tag which seems unintended.
> 2. Thoughts on "Edit Profile" -> "Manage Account" since not all of the edits on this page are publicly facing.
> 3. We have a route on the website `/logout` but no links to it anywhere so far. I think this page makes the most sense to include that link since it wouldn't be used too commonly by most users.
> 4. I think we have something similar on some other form components but it would help a lot to highlight which fields have been edited and are currently unsaved so I can know before clicking the cancel or save button and making unintended edits.

Second Faris's fourth point about which fields have been edited. Right now it's a little confusing when changes are actually updated. Certain fields like profile picture and resume are updated at the moment you upload while other fields are only saved when you click save at the bottom. I think it's also a little confusing that pressing enter in one field will attempt to save all of them, including the password. Maybe we should think about moving the reset password to it's own section. Also, editing the email address is a pretty big change since it require's reverifying your email and I wonder if people can get locked out of their accounts by changing it, so maybe that field also needs more warning. Perhaps we can take inspiration from the v1 profile edit page where each section, separated by lines, gets saved together.

However, I think we should go ahead and merge this branch and then address those issues in a future PR.

SheepTester:
> 3\. ~Edit: I can't seem to see my current resume on the page~ Editedit: I deleted my cookie and it's fine, dunno what was up with that.

Your cookie caches your user data locally, so it can get stale when you make changes to your profile (such as earning points) from a different browser or device. We probably should look into better ways to store the user object

src/components/admin/event/EventDetailsForm/index.tsx
SheepTester: > Does `Cropper` block form submission if the cover is too large? if not, then we should probably enforce a check here before allowing submission

yes, but I can add another check inside `createNewEvent` as a sanity check just in case

raymosun (approved):

farisashai (approved):
I think the reset password form should be separate visually and with its own action button since it's destructive and not the same as updating profile fields. aside from that, this should be fine to merge and iterate later

PR #94, opened 2023-11-28
Andy/store item page

### LINKED ISSUE #56: Create Store Item Page
Create the store item page at `/store/item/[item-uuid]` based on these [figma designs](https://www.figma.com/file/cgbeW6vrqgxW0i1vpbUn1C/Store-(Redesign-2023)?type=design&node-id=654-2155&mode=design&t=jCa9Nm6GIE583hHg-0) (check the 'ready for dev' sections).

Requires #54.

Encompasses
- Desktop layout for item page
- Mobile layout for item page
- Modal for adding item to cart

image

WishingWell13:
# Info

Closes #56

# Description

motivation or context.

Create the store item page at /store/item/[item-uuid] based on these [figma designs](https://www.figma.com/file/cgbeW6vrqgxW0i1vpbUn1C/Store-(Redesign-2023)?type=design&node-id=654-2155&mode=design&t=jCa9Nm6GIE583hHg-0) (check the 'ready for dev' sections).

Requires https://github.com/acmucsd/membership-portal-ui-v2/issues/54.

Encompasses

- Desktop layout for item page

To Do:
- Mobile layout for item page 
- Modal for adding item to cart (Waiting for @SheepTester PR #91)
- Carousel for images

## Changes

- Designed Desktop item page under `src\pages\store\item\[uuid].tsx` 

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.
- [x] I have run and passed all new and existing Cypress tests. Add screenshots below.
image

# Checklist

      anywhere else.

# Screenshots

image

image

image

image


src/components/common/ThemeToggle/style.module.scss
```diff
@@ -1,3 +1,7 @@
+@use 'src/styles/vars.scss' as vars;
+// EXAMPLE: vars.$light-primary-1 (see SizeSelector p{} styling)
+// TODO: Update text size, color, background color, border color, etc. to follow vars presets
```
farisashai: Does this have anything to do with the store item page? Otherwise I would remove and just open a new issue for it

src/styles/pages/store/item.module.scss
```diff
@@ -0,0 +1,18 @@
+.container {
+  display: flex;
+  flex-direction: column;
+  gap: 16px;
```
farisashai:
```suggestion
  gap: 1rem;
```

src/styles/pages/store/item.module.scss
```diff
@@ -0,0 +1,18 @@
...
+  align-items: center;
+  display: flex;
+  flex-direction: row;
+  gap: 79px;
+}
+
+.navbarBodyDiv {
+  display: flex;
+  flex-direction: column;
+  gap: 32px;
```
farisashai:
```suggestion
  gap: 2rem;
```

src/styles/pages/store/item.module.scss
```diff
@@ -0,0 +1,18 @@
...
+  display: flex;
+  flex-direction: column;
+  gap: 16px;
+}
+
+.rowContainer {
+  align-items: center;
+  display: flex;
+  flex-direction: row;
+  gap: 79px;
```
farisashai:
```suggestion
  gap: 5rem;
```

src/lib/api/StoreAPI.ts
```diff
@@ -1,14 +1,38 @@
 import { config } from '@/lib';
+import { UUID } from '@/lib/types';
 import type {
```
farisashai, line 3:
```suggestion
import type { UUID } from '@/lib/types';
```

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const validText = inStock ? (
+    <p className={styles.valid}>This item is in stock.</p>
+  ) : (
+    <p className={styles.error}>This item is out of stock.</p>
+  );
+
+  const buttonStyle = inStock ? styles.buttonInStock : styles.buttonNoStock;
+
+  const optionArr: Array<{ value: string; label: string } | '---'> = [];
```
farisashai: one line conditions like this and the previous variable should just be inlined directly into the html where they're used to reduce clutter

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  for (let i = 1; i <= maxCanBuy; i += 1) {
+    optionArr.push({ value: `${i}`, label: `${i}` });
+  }
+
```
farisashai: does this value need to be a string type or can we keep it as a plain number?

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <input
+          className={`${buttonStyle} ${styles.button}`}
+          type="button"
+          value={inStock ? 'Add to Cart' : 'Out of Stock'}
+          onClick={() => {
+            setInCart(!inCart);
+          }}
+          disabled={!(inStock || maxCanBuy > 0)}
+        />
+      </div>
```
farisashai:
```suggestion
          disabled={!inStock || maxCanBuy === 0}
```
I don't think this is logically valid in every situation but someone can correct me if I'm wrong

src/components/store/ItemHeader/style.module.scss
```diff
@@ -0,0 +1,59 @@
...
+  flex-direction: row;
+  gap: 8px;
+}
+
+.itemHeaderGroup {
+  display: flex;
+  flex-direction: column;
+  gap: 8px;
+  //   overflow: auto;
+  // padding: 5px 3px;
```
farisashai: delete comments

farisashai (changes requested):
haven't looked at a preview but some quick initial nitpicks for code quality

src/components/common/ThemeToggle/style.module.scss
WishingWell13: Nope, is a marker to help Diamond Staff to get started with this issue: https://github.com/acmucsd/membership-portal-ui-v2/issues/95

Would it help the Diamond Staff / OK to keep it until then and have them delete it?

src/components/store/AddCartButton/index.tsx
WishingWell13: Yes, it needs to be strings. The `Dropdown` component only takes in strings. 
image
image

src/components/store/AddCartButton/index.tsx
WishingWell13: Can you elaborate on what you mean?

src/styles/pages/store/item.module.scss
farisashai, line 1: i would rename this file to `src/styles/pages/StoreItemPage.module.scss`. theres no point to nesting the css in folders really at this level

src/styles/pages/store/item.module.scss
```diff
@@ -0,0 +1,18 @@
+.container {
+  display: flex;
+  flex-direction: column;
+  gap: 1rem;
+}
+
+.rowContainer {
+  align-items: center;
```
farisashai:
```suggestion
  align-items: flex-start;
```

src/styles/pages/store/item.module.scss
```diff
@@ -0,0 +1,18 @@
...
+  display: flex;
+  flex-direction: row;
+  gap: 5rem;
```
farisashai:
```suggestion
  gap: 5rem;
  justify-content: center;
```

.vscode/settings.json
```diff
@@ -3,8 +3,8 @@
   "editor.formatOnSave": true,
   "typescript.preferences.importModuleSpecifier": "non-relative",
   "editor.codeActionsOnSave": {
-    "source.fixAll": true,
-    "source.organizeImports": true
+    "source.fixAll": "explicit",
+    "source.organizeImports": "explicit"
```
farisashai:
```suggestion
    "source.fixAll": true,
    "source.organizeImports": true
```

package.json
```diff
@@ -22,6 +22,9 @@
     "test": "yarn cypress run",
     "test:ui": "yarn cypress open"
   },
+  "resolutions": {
+    "@types/react": "18.0.21"
```
farisashai: seeking to learn: what is this for?

src/components/common/ThemeToggle/style.module.scss
farisashai: I think it's simpler to track with a Github issue than have to come look at a comment in the repo to remember this

src/components/layout/Navbar/index.tsx
```diff
@@ -125,7 +125,11 @@ const Navbar = ({ user }: NavbarProps) => {
           <ProfileIcon className={styles.iconLink} />
           Profile
         </Link>
-        <Link className={styles.mobileNavItem} href={config.storeRoute}>
+        <Link
+          onClick={() => setMenuOpen(false)}
```
farisashai: why is this onClick here for a link?

.vscode/settings.json
SheepTester: what does this do

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+}: AddCartButtonProps) => {
+  const [mounted, setMounted] = useState(false);
+  const [amount, setAmount] = useState<number>(1);
+
+  const myID = useId();
+
+  const maxCanBuy = Math.min(lifetimeRemaining, montlyRemaining);
+
+  useEffect(() => {
+    // console.log(`currSize ${currSize} inStock ${inStock} inCart ${inCart}`);
```
farisashai: delete unused comments

a tip for future PRs you open - try to do a full self review before you have anyone else look at it and if you get a comment, see everywhere it applies and fix it so you don't need to read the same comment written multiple times

src/components/layout/Navbar/index.tsx
SheepTester: this used to be consistent, but another PR moved this `onClick` handler to the parent element, so for consistency you can now remove this line

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <p className={styles.valid}>This item is in stock.</p>
+  ) : (
+    <p className={styles.error}>This item is out of stock.</p>
+  );
+
+  const optionArr: Array<{ value: string; label: string } | '---'> = [];
+
+  for (let i = 1; i <= maxCanBuy; i += 1) {
+    optionArr.push({ value: `${i}`, label: `${i}` });
+  }
```
farisashai: you need to add a check if maxCanBuy === 0 then push { value: '0', label: '0' } so the dropdown renders correctly still

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+  optionArr.push('---');
+
+  return (
+    <div className={styles.addCartGroup}>
+      {maxCanBuy > 0 ? (
+        <p>You can buy up to {maxCanBuy} of this item.</p>
+      ) : (
+        <p>You can&apos;t buy any more of this item!.</p>
+      )}
+      <p>{`In cart: ${inCart}`}</p>
```
farisashai: i think this will work but please confirm
```suggestion
      <p>In cart: {inCart}</p>
```

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+          />
+        </div>
+
+        <input
+          className={`${inStock ? styles.buttonInStock : styles.buttonNoStock} ${styles.button}`}
+          type="button"
+          value={inStock ? 'Add to Cart' : 'Out of Stock'}
+          onClick={() => {
+            setInCart(!inCart);
+          }}
```
SheepTester: It looks like this button toggles the item being in the cart or not. You should update the button's label so it says "remove" when it's already in the cart

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <div className={styles.quantityColumn}>
+          <h4>Quantity</h4>
+          <Dropdown
+            name={`options${myID}`}
+            ariaLabel={`Dropdown to select the number of items to purchase for ${myID}`}
+            options={optionArr}
+            value={`${amount}`}
+            onChange={val => setAmount(Number(val))}
+          />
+        </div>
```
SheepTester: the dropdown is empty when `maxCanBuy` is zero; you should just hide it when you can't select the quantity

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+interface AddCartButtonProps {
+  currSize: string | undefined;
+  // Justification for disabling rules: This seems to be a false positive.
+  // https://stackoverflow.com/q/63767199/
+  // eslint-disable-next-line no-unused-vars
+  inStock: boolean;
+  inCart: boolean;
+  lifetimeRemaining: number;
+  montlyRemaining: number;
```
SheepTester: typo? probably meant "monthly"

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const maxCanBuy = Math.min(lifetimeRemaining, montlyRemaining);
+
+  useEffect(() => {
+    // console.log(`currSize ${currSize} inStock ${inStock} inCart ${inCart}`);
+    setMounted(true);
+  }, []);
+
+  if (!mounted) {
+    return null;
+  }
```
SheepTester: Is there a reason why you don't want it to mount on the server side?

src/pages/store/item/[uuid].tsx
```diff
@@ -0,0 +1,88 @@
...
+      <Navbar balance={credits} showBack />
+      <div className={styles.rowContainer}>
+        <Image
+          src={getDefaultMerchItemPhoto(item) || ''}
+          alt="Picture of item "
+          width={600}
+          height={600}
+        />
+        <div className={styles.container}>
+          <h1>Store Item Page {size}</h1>
```
Comment (farisashai) [inline src/pages/store/item/[uuid].tsx]:
```suggestion
```
i don't think this is necessary

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+interface AddCartButtonProps {
+  currSize: string | undefined;
+  // Justification for disabling rules: This seems to be a false positive.
+  // https://stackoverflow.com/q/63767199/
+  // eslint-disable-next-line no-unused-vars
+  inStock: boolean;
+  inCart: boolean;
+  lifetimeRemaining: number;
+  montlyRemaining: number;
+  setInCart: (value: boolean) => void;
```
SheepTester: in React's naming conventions, starting something with `set` should be reserved for something from `useState` within a component. if you want to pass the new `inCart` value back up to the parent component, the prop should start with `on`, like an event handler

```suggestion
  onCartChange: (inCart: boolean) => void;
```

src/components/store/AddCartButton/index.tsx
farisashai: your boolean condition `!(inStock || maxCanBuy > 0)` expands to `!inStock && !(maxCanBuy > 0)` or `!inStock && maxCanBuy == 0` using demorgans law so it will only disable if the item is not in stock and it cannot be bought but we want it to disable if either of these are true

src/components/store/AddCartButton/style.module.scss
```diff
@@ -0,0 +1,62 @@
...
+$indicator-size: 30px;
+$switch-start: 15%;
+$button-roundness: 10px;
+
+.addCartGroup {
+  display: flex;
+  flex-direction: column;
+  gap: 1rem;
+  overflow: visible;
+  // padding: 5px 3px;
```
farisashai: delete

src/components/store/ItemHeader/index.tsx
```diff
@@ -0,0 +1,37 @@
+import BlueDiamond from '@/public/assets/icons/blue-diamond.svg';
+import { useEffect, useState } from 'react';
+import styles from './style.module.scss';
+
+interface ItemHeaderProps {
+  itemName: string;
+  cost: number | undefined;
+  // Justification for disabling rules: This seems to be a false positive.
+  // https://stackoverflow.com/q/63767199/
+  // eslint-disable-next-line no-unused-vars
```
farisashai: what is this comment for? can you double check your own diff especially when you copy paste code in or delete something that might have been relevant

src/components/store/AddCartButton/index.tsx
SheepTester: not sure, but I think he's just not confident his suggestion is 100% correct either

your current expression says that you can add to cart if it's in stock, but if it's not, you can still add to cart if `maxCanBuy > 0`. so currently, you can get a weird situation like this:

image

> You can't buy any more of this item!.

but then it also says

> This item is in stock.

and then it lets you add to cart

src/components/store/ItemHeader/style.module.scss
```diff
@@ -0,0 +1,57 @@
...
+    border-width: 1px;
+    font-size: vars.$input-text-medium;
+    font-weight: bold;
+    height: 40px;
+    width: 200px;
+  }
+
+  .buttonInStock {
+    background-color: var(--theme-primary-2);
+    color: vars.$light-background; // or var(--theme-background);
```
farisashai: does this work on light and dark mode?

src/components/store/ItemHeader/style.module.scss
```diff
@@ -0,0 +1,57 @@
...
+    color: vars.$danger-1;
+  }
+
+  .valid {
+    color: vars.$success-1;
+  }
+
+  .button {
+    border-color: var(--theme-accent-line-2);
+    border-radius: $button-roundness;
```
farisashai: variables are unnecessary if they're only used once you can write the value directly in here

src/components/store/ItemHeader/index.tsx
SheepTester: I don't think you need this here

src/components/store/SizeSelector/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+          <form className={styles.switch}>{myOptions}</form>
+        </>
+      )}
+    </div>
+  );
+};
+export default SizeSelector;
+
+// Links for knowledge:
+// https://medium.com/@vincent.bocquet/typescript-with-react-usestate-hook-f701309384a
```
farisashai: can you delete this comment everywhere or leave it once at max - it seems like its in a lot of files that have been copy pasted

src/components/store/ItemHeader/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+const ItemHeader = ({ itemName, cost }: ItemHeaderProps) => {
+  const [mounted, setMounted] = useState(false);
+
+  useEffect(() => {
+    setMounted(true);
+  }, []);
+
+  if (!mounted) {
+    return null;
+  }
```
SheepTester: I don't think this is necessary for this component

`useEffect` hurts performance because it makes React render once, run the `useEffect`, then render again. it also prevents the component from being rendered on the server side, which negates the benefits of using NextJS. using `useEffect` is often necessary, but you'd want to avoid it if possible

src/components/store/SizeSelector/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+          onClick={() => setSize(val.metadata?.value)}
+        />
+        <label htmlFor={tempId}>{val.metadata?.value}</label>
+      </>
+    );
+  });
+
+  return (
+    <div className={styles.sizeSelector}>
+      {options.length > 1 && (
```
farisashai: it's safer to not use conditional rendering like this and replace `condition && <Component />` with `condition ? <Component /> : null` because in the former case, you can get weird rendering bugs where if condition evaluates to 0 or some false values, it will still render that variable's value

src/components/store/SizeSelector/style.module.scss
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  .switch {
+    align-items: center;
+    border-color: #9f9f9f;
+    border-radius: $component-roundness;
+    display: flex;
+    float: left;
+    gap: 12px;
+    // border-style: solid;
```
farisashai: delete these

src/components/store/SizeSelector/style.module.scss
```diff
@@ -0,0 +1,73 @@
+@use 'src/styles/vars.scss' as vars;
+$indicator-size: 30px;
+$switch-gap: 127%;
+$switch-gap-num: calc($switch-gap / 100%);
```
farisashai: delete any of these variables that you haven't used at least twice

src/components/store/ItemHeader/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  return (
+    <div className={styles.itemHeaderGroup}>
+      <h1>{itemName}</h1>
+      <div className={styles.price}>
+        <p>{cost || 'Select Size Option First'}</p>
+        <BlueDiamond />
+      </div>
```
SheepTester: There's a `Diamonds` component that does this already (see the `ItemCard` component to see how it's used), so you don't need blue-diamond.svg

I wonder if it's possible for there to be free items on the store. In that case, it would show "Select Size Option First" instead of "0 diamonds." Maybe you should check to make sure `cost !== undefined` instead of checking if it's truthy, just in case

src/lib/api/StoreAPI.ts
farisashai: why is this resolved? it doesn't look like it was changed

farisashai (changes requested):
left comments for everything except the file for the actual store page, i'll take a closer look there when the page is fully done and matches the given design on desktop/mobile (we can skip the image carousel for now and anything that would take too long)

src/components/store/SizeSelector/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }, []);
+
+  if (!mounted) {
+    return null;
+  }
+
+  const myOptions = options.map(val => {
+    const tempId = val.metadata?.value ? `${val.metadata.value}${myID}` : `${val.uuid}${myID}`;
+    return (
+      <>
```
SheepTester: Hmm, does React not give a warning about this missing a `key` prop? React usually requires a `key` prop when you return a list of elements, like here, so it can keep track of which is which

You can't add a prop directly to `<></>` though. You'd have to import `Fragment` (which is what `<>` is short for) from React then do

```suggestion
      <Fragment key={val.uuid}>
```

src/components/store/SizeSelector/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  options: PublicMerchItemOption[];
+  // Justification for disabling rules: This seems to be a false positive.
+  // https://stackoverflow.com/q/63767199/
+  // eslint-disable-next-line no-unused-vars
+  setSize: (value: string) => void;
+}
+
+const SizeSelector = ({ currSize, options, setSize }: SizeSelectorProps) => {
+  const [mounted, setMounted] = useState(false);
+  const myID = `${useId()}`;
```
SheepTester: I think `useId` already returns a string so you don't need to cast it into a string again

src/pages/store/item/[uuid].tsx
```diff
@@ -0,0 +1,88 @@
...
+  item: PublicMerchItemWithPurchaseLimits;
+}
+const StoreItemPage = ({ user: { credits }, item }: ItemPageProps) => {
+  const [size, setSize] = useState<string | undefined>(item.options.length <= 1 ? 'Y' : undefined);
+  const [inCart, setInCart] = useState<boolean>(false);
+
+  const currOption =
+    item.options.length <= 1
+      ? item.options[0]
+      : item.options[item.options.findIndex(val => val.metadata?.value === size)];
```
Comment (SheepTester) [inline src/pages/store/item/[uuid].tsx]: you can also use `.find`

```suggestion
      : item.options.find(val => val.metadata?.value === size);
```

src/styles/pages/store/item.module.scss
WishingWell13: Should I also rename the other files in that folder, or leave that for another PR?

farisashai: there's just one file so might as well

src/pages/store/item/[uuid].tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  return (
+    <div className={styles.navbarBodyDiv}>
+      <Navbar balance={credits} showBack />
+      <div className={styles.rowContainer}>
+        <Image
+          src={getDefaultMerchItemPhoto(item) || ''}
```
Comment (SheepTester) [inline src/pages/store/item/[uuid].tsx]: I don't think `getDefaultMerchItemPhoto` returns undefined anymore, so you can get rid of the `|| ''`

```suggestion
          src={getDefaultMerchItemPhoto(item)}
```

src/pages/store/item/[uuid].tsx
```diff
@@ -0,0 +1,88 @@
...
+          alt="Picture of item "
```
Comment (SheepTester) [inline src/pages/store/item/[uuid].tsx]: trailing space; you could also make the description a bit more specific, using `item.itemName` or something, like

```suggestion
          alt={`Picture of ${item.itemName}`}
```

package.json
WishingWell13: "Adding resolutions makes sure it uses the correct version of types/react"

See here: https://discord.com/channels/573028991527550986/1187165206305046670/1191216173329694770
image

SheepTester (commented):
I know it says "please select a size" like twice, but my eyes still immediately went towards the "Out of stock" button

image

I think it should hide things like the add to cart button until you select a size, like

image

.vscode/settings.json
WishingWell13: It's been automatically switching to that. Unsure why. I must have missed reverting it once. I'll change it back.

src/components/store/ItemHeader/style.module.scss
SheepTester: because the button background color is blue, white text color is legible in both dark and light mode. but for consistency with the login page's button, I think it should use `var(--theme-text-on-background1)` or whatever it's called, so the button text is black in dark mode

src/components/store/AddCartButton/index.tsx
SheepTester: also, the default border is showing on the button

image

we do remove the border by default, but only for `<button>` elements. I think you should change `<input type="button">` to `<button type="button">` for consistency with the rest of the codebase. it would fix the border

src/components/store/SizeSelector/index.tsx
WishingWell13: image

WishingWell13: React was not giving me an error for this, which is odd.

src/components/store/AddCartButton/index.tsx
WishingWell13: https://www.npmjs.com/package/next-themes

I thought it would mess up the dark/light mode, but it seems like it won't since we don't use `useTheme()` specifically in this file. Removed it.

src/components/layout/Navbar/index.tsx
SheepTester: it's to close the menu after clicking on a menu item for the mobile navbar. this line is no longer needed though

src/components/store/AddCartButton/index.tsx
WishingWell13: image

src/components/store/SizeSelector/index.tsx
SheepTester: you only need it if you're returning an array of elements. for the `Fragment` in your screenshot, it's not in an array, so you don't need a `key`

src/components/store/ItemHeader/index.tsx
WishingWell13: image

Originally found in the `Dropdown` component. The placement of the comment was off, fixed in the new update.

Not quite sure what it is. Sent a message in discord.

src/components/store/ItemHeader/index.tsx
WishingWell13: Removed

src/components/store/ItemHeader/style.module.scss
WishingWell13: Which variables are you referring to? I see them be used multiple times:

image


image

and headings are standard with the Figma.

src/components/store/ItemHeader/style.module.scss
WishingWell13: image

Figma says use the `background` color I believe. If there is no functional difference I am good with keeping `var(--theme-text-on-background1)`

src/components/store/AddCartButton/index.tsx
WishingWell13: If someone could show me how to get to the view y'all are using to make small commits, that would be helpful for this.

src/components/store/AddCartButton/index.tsx
WishingWell13: Should I also disable the purchase button if the user selects `0` as the amount to buy?

src/components/store/ItemHeader/style.module.scss
WishingWell13: Saw the other comment, resolved.

src/components/layout/Navbar/index.tsx
WishingWell13: Removed `onClick` only if I understood that right

src/components/store/AddCartButton/index.tsx
WishingWell13: Need a method to test this

Don't see an item that has hit a lifetime limit and is in stock on the testing portal
https://testing.members.acmucsd.com/store

src/components/store/ItemHeader/index.tsx
WishingWell13: If cost is undefined what would I display?

src/components/store/SizeSelector/index.tsx
WishingWell13: Which lines were you referring to then? I think I'm misunderstanding.

src/lib/api/StoreAPI.ts
WishingWell13: image

Can you send me a screenshot of your end? Looks good on mine.

src/components/store/AddCartButton/index.tsx
WishingWell13: This path using the acm@ucsd.edu email: `store/item/18d76c1d-0d77-40a5-81cf-a7438e89a117`

image

src/components/common/ThemeToggle/style.module.scss
WishingWell13: Already created an issue here: https://github.com/acmucsd/membership-portal-ui-v2/issues/95

I figured that having it directly in the codebase might give any Diamond Staff who are unfamiliar with the issue system a bridge between "finding the files only through issues" and "having an indicator on the file".

If we still don't like the idea we can delete it.

src/components/store/AddCartButton/index.tsx
WishingWell13: image
image

The Figma has a border on the button. Should I put the border back?

Without a border it looks like this:
https://github.com/acmucsd/membership-portal-ui-v2/pull/94/files/385c9dfde91b12565c53d13580c374e954e46afd#r1443356378

src/components/store/AddCartButton/index.tsx
SheepTester: not really a special view or anything, but reviewing all the changes you made https://github.com/acmucsd/membership-portal-ui-v2/pull/94/files can be helpful (albeit it can be a lot to look over), because these are the changes we are reviewing, and it highlights the lines you made. so you could use this to find all the comments you added, and see if they're unused or not

for seeing if a suggestion can be applied elsewhere in your code, you can go back into VS Code and search across all the files (ctrl+shift+F). for example, you could use this to look up some kinds of comments, like `// Justification for disabling rules`, to see if they're necessary or not

for the little suggested commits we make in reviews, we do that by doing

````md
```suggestion
suggested change here
```
````

and then github will show a button to make the commit

File: src/components/store/AddCartButton/index.tsx (hunk 18 above)
Comment (SheepTester) [reply src/components/store/AddCartButton/index.tsx]: hmm, I think you should keep the border when the button is greyed out. I think the button looks better without the border when it's blue, but it's up to you whether the border on the blue button looks nicer

File: src/components/store/ItemHeader/style.module.scss (hunk 26 above)
Comment (SheepTester) [reply src/components/store/ItemHeader/style.module.scss]: oh, I think `var(--theme-background)` might be better. I mixed them up, oops

`var(--theme-text-on-background1)` would be black in light mode, white in dark mode (used for text), while `var(--theme-background)`  is the opposite: white in light mode, black in dark mode (used for the page background color)

the button's blue background is *brighter* in dark mode, so the text color should be darker to compensate, so I think using `var(--theme-background)` for the button text would be better

it's kind of confusing; here's the sign in button for reference:

image image

File: src/components/store/SizeSelector/index.tsx (hunk 34 above)
Comment (SheepTester) [reply src/components/store/SizeSelector/index.tsx]: the lines inside `const myOptions`

File: public/assets/icons/blue-diamond.svg
Comment (SheepTester) [inline public/assets/icons/blue-diamond.svg:1]: since you're using the `Diamonds` component now, I don't think you need this file anymore

File: src/components/store/AddCartButton/index.tsx [hunk 39]
```diff
@@ -0,0 +1,108 @@
...
+import styles from './style.module.scss';
+
+interface AddCartButtonProps {
+  currSize: string | undefined;
+  inStock: boolean;
+  inCart: boolean;
+  lifetimeRemaining: number;
+  monthlyRemaining: number;
+  amountToBuy: number;
+  // eslint-disable-next-line no-unused-vars
```
SheepTester: by the way, since I merged my PR #91, I fixed the issue where it thinks the parameter is unused. if you merge in changes from `main`, you can find and remove all `// eslint-disable-next-line no-unused-vars` and it shouldn't give a warning anymore

src/pages/store/item/[uuid].tsx
```diff
@@ -0,0 +1,95 @@
+import { Navbar } from '@/components/store';
+import AddCartButton from '@/components/store/AddCartButton';
+import ItemHeader from '@/components/store/ItemHeader';
+import SizeSelector from '@/components/store/SizeSelector';
```
Comment (farisashai) [inline src/pages/store/item/[uuid].tsx]: all 3 of these can be imported together in one line from the parent folder with the Navbar

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+              />
+            </div>
+          )}
+
+          <button
+            className={`${inStock && maxCanBuy > 0 ? styles.buttonInStock : styles.buttonNoStock} ${
+              styles.button
+            }`}
+            type="button"
+            title={`${buyButtonText} Button`}
```
SheepTester: I don't think you need the `title` attribute since the button already has a label visible

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+            value={buyButtonText}
+            onClick={() => {
+              onCartChange(!inCart);
+            }}
+            disabled={!inStock || maxCanBuy === 0}
+          >
+            {' '}
+            {buyButtonText}{' '}
```
SheepTester: out of curiosity, do you need these spaces here? if not, then might as well remove them; `{' '}` is kind of unsightly

```suggestion
            {buyButtonText}
```

src/components/store/ItemHeader/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+  cost: number | undefined;
+}
+
+const ItemHeader = ({ itemName, cost }: ItemHeaderProps) => {
+  return (
+    <div className={styles.itemHeaderGroup}>
+      <h1>{itemName}</h1>
+      <div className={styles.price}>
+        {cost === undefined ? <p>Select Size Option First For Price</p> : <Diamonds count={cost} />}
+      </div>
```
SheepTester: I think you should use sentence case for the no size selected message, "Select size option first for price"

Alternatively, you could just hide the price silently if no size is selected, since it already says "Please select a size" in red below:
image

src/components/store/SizeSelector/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const [mounted, setMounted] = useState(false);
+  const myID = useId();
+
+  useEffect(() => {
+    setMounted(true);
+  }, []);
+
+  if (!mounted) {
+    return null;
+  }
```
SheepTester: I don't believe you need this for this component either

src/lib/api/StoreAPI.ts
SheepTester: I think he just wants you to put `type` between `import` and `{ UUID }` on line 2, because we're only using `UUID` as a type in that file

SheepTester (commented):
very good, glad to see the dropdown being hidden when you can only buy one of the item

src/components/common/ThemeToggle/style.module.scss
farisashai: let's delete all of these comments since the feature has been completed already

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+          {!inStock || maxCanBuy <= 1 ? null : (
+            <div className={styles.quantityColumn}>
+              <h4>Quantity</h4>
+              <Dropdown
+                name={`options${myID}`}
+                ariaLabel={`Dropdown to select the number of items to purchase for ${myID}`}
+                options={optionArr}
+                value={`${amountToBuy}`}
+                onChange={val => onAmountChange(Number(val))}
+              />
```
SheepTester: Is there a reason why you're using a dropdown instead of a number input for quantity?

src/components/store/SizeSelector/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <div className={styles.sizeSelector}>
+      {options.length > 1 ? (
+        <Fragment key={uuid}>
+          <h4>Size</h4>
+          <form className={styles.switch}>{myOptions}</form>
+        </Fragment>
+      ) : null}
+    </div>
+  );
```
SheepTester: What I meant for adding the `key` prop was to add it to elements inside arrays, like `myOptions`, so React can tell the items apart. So below in the suggestion, I added `key` to the `Fragment` inside `map` since it creates an array.

```suggestion
  const myOptions = options.map(val => {
    const tempId = val.metadata?.value ? `${val.metadata.value}${myID}` : `${val.uuid}${myID}`;
    return (
      <Fragment key={tempId}>
        <input
          id={tempId}
          name="state-d"
          type="radio"
          defaultChecked={currSize === val.metadata?.value}
          onClick={() => onSizeChange(val.metadata?.value)}
        />
        <label htmlFor={tempId}>{val.metadata?.value}</label>
      </Fragment>
    );
  });

  return (
    <div className={styles.sizeSelector}>
      {options.length > 1 ? (
        <>
          <h4>Size</h4>
          <form className={styles.switch}>{myOptions}</form>
        </>
      ) : null}
    </div>
  );
```

The fragment at the bottom doesn't need a `key` because React knows that there will it either be one element or none

src/components/store/AddCartButton/index.tsx
SheepTester: currently it crashes the browser and server when the max quantity is large: https://membership-portal-ui-v2-git-andy-storeitempage-acmucsd.vercel.app/store/item/a174f5b2-de6c-44f2-b99c-f36a27221797

package.json
```diff
@@ -23,6 +23,9 @@
...
```
alexzhang1618: Wondering if we still need this since you've switched computers 😁

alexzhang1618 (changes requested):
Some visual nitpicks:

I don't like how the image can take up variable space on the left column here, I think we should keep the ratio standard and scale the image to fit inside of that.
Screenshot 2024-01-22 at 10 00 34 PM
On this one, the variant isn't fitting inside of the variant button
Screenshot 2024-01-22 at 10 01 02 PM

Additionally, it seems like the variant buttons and some of the fonts/spacings aren't consistent with Figma. For reference: 
image
image

src/components/store/SizeSelector/style.module.scss
```diff
@@ -0,0 +1,64 @@
...
+      width: #{$indicator-size * 2};
+    }
+
+    input {
+      opacity: 0;
+      position: absolute;
+    }
+
+    input:checked {
+      border-color: var(--size-selector-border);
```
SheepTester: this border shouldn't be noticeable since it has zero opacity, right?

src/components/store/SizeSelector/style.module.scss
```diff
@@ -0,0 +1,64 @@
...
+  gap: 0.75rem;
+  overflow: auto;
+
+  h4 {
+    color: var(--theme-text-on-background-1);
+    font-size: vars.$general-heading-3-4;
+    font-weight: bold;
+  }
+
+  .switch {
```
SheepTester, line 11: image

you should change `width` to `min-width` for the labels (and add some horizontal padding) and add `flex-wrap: wrap;` to `.switch` so if there's not enough room, the options go onto the next line

image

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  if (maxCanBuy === 0) {
+    optionArr.push({ value: `0`, label: `0` });
+  }
+
+  for (let i = 1; i <= maxCanBuy; i += 1) {
+    optionArr.push({ value: `${i}`, label: `${i}` });
+  }
+
+  optionArr.push('----');
```
SheepTester: What is the purpose of the final line?

image

src/styles/pages/StoreItemPage.module.scss
```diff
@@ -0,0 +1,44 @@
...
+.rowContainer {
+  align-items: flex-start;
+  align-self: center;
+  display: flex;
+  flex-direction: row;
+  gap: 5rem;
+  justify-content: center;
+  width: 95%;
+
+  @media (max-width: 600px) {
```
SheepTester: I think the image can get pretty small before converting into a single column layout. I would increase this `max-width` so the single column layout kicks in earlier

image

This codebase would prefer you use the media breakpoint variables in vars.scss for consistency,  rather than hard-coded values like `600px`. You can search for `@media` in other files to see how they're used

src/styles/themes.scss
```diff
@@ -91,6 +92,7 @@
 
   --theme-mode-indicator: #{vars.$dark-mode-indicator};
   --theme-mode-icon: #{vars.$white};
+  --size-selector-border: #{vars.$dark-primary-4};
```
SheepTester: I think for consistency, since all the other variables start with `--theme-`, this one should too

SheepTester (changes requested):
Very excellent, just some nitpicks.

This is just a quick design review; I'll do a code review later

- I think you should put "You can buy up to 20 of this item." next to the quantity dropdown
- I think it might not be necessary to say "Item Description." Also, maybe it might look nicer if the description is right under the title/price
- We probably don't have to say "This item is in stock." since the fact that you can press on the "Add to Cart" button says that already

image

and here, it's a bit redundant saying "Out of stock" or "Limit reached" twice

image

image

I think just having the disabled button is enough



also, you might want to add `word-break: break-word;` for extreme cases like this:

image

https://membership-portal-ui-v2-git-andy-storeitempage-acmucsd.vercel.app/store/item/0702252f-a50b-4948-ad21-e1ca38850413

though admittedly, it probably won't happen in practice

src/pages/store/item/[uuid].tsx
```diff
@@ -0,0 +1,93 @@
...
+            onCartChange={setInCart}
+            currSize={size}
+            inStock={currOption?.quantity != null && currOption?.quantity >= 1}
+            lifetimeRemaining={item.lifetimeRemaining}
+            monthlyRemaining={item.monthlyRemaining}
+            amountToBuy={amount}
+            onAmountChange={setAmount}
+          />
+          <h4>Item Description</h4>
+          <p>{item.description}</p>
```
Comment (SheepTester) [inline src/pages/store/item/[uuid].tsx]: image

You should add `white-space: pre-wrap;` so newline characters are kept. You might also want to increase `line-height` to, say, `1.5`, so the text doesn't look as dense:

image

.vscode/settings.json
farisashai, line 1: lets try to make sure to revert any changes in the PR's diff that aren't related to the feature you're working on

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+          <button
+            className={`${inStock && maxCanBuy > 0 ? styles.buttonInStock : styles.buttonNoStock} ${
+              styles.button
+            }`}
+            type="button"
+            title={`${buyButtonText} Button`}
+            value={buyButtonText}
+            onClick={() => {
+              onCartChange(!inCart);
+            }}
```
farisashai:
```suggestion
            onClick={() => onCartChange(!inCart) }
```

you can inline single line arrow functions

src/components/store/ItemHeader/style.module.scss
```diff
@@ -0,0 +1,42 @@
...
+}
+
+.itemHeaderGroup {
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+  gap: 0.9rem;
+
+  h1 {
+    @media (max-width: 600px) {
```
farisashai: nit: hardcoded media query

src/components/store/ItemHeader/style.module.scss
```diff
@@ -0,0 +1,42 @@
...
+  display: flex;
+  flex-direction: row;
+  gap: 0.5rem;
+}
+
+.itemHeaderGroup {
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+  gap: 0.9rem;
```
farisashai: super tiny nit: lets use a consistent scale for padding values of values like 0.5rem, 0.75, 1rem, 1.5, 2rem to avoid uneven values everywhere

src/components/store/SizeSelector/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <label htmlFor={tempId}>{val.metadata?.value}</label>
+      </Fragment>
+    );
+  });
+
+  return (
+    <div className={styles.sizeSelector}>
+      {options.length > 1 ? (
+        <>
+          <h4>Size</h4>
```
farisashai: i'm not sure if this data comes from the api but i'm pretty sure that store admins can define any subset of item types here - for example, sticker options such as color: red, blue, pink.

if there's a label for these options, we should use that otherwise we need to make this text more generic to encompass any set of variants/options

src/pages/store/item/[uuid].tsx
```diff
@@ -0,0 +1,93 @@
...
+import Image from 'next/image';
+import { useState } from 'react';
+
+interface ItemPageProps {
+  user: PrivateProfile;
+
+  item: PublicMerchItemWithPurchaseLimits;
+}
+const StoreItemPage = ({ user: { credits }, item }: ItemPageProps) => {
+  const [size, setSize] = useState<string | undefined>(item.options.length <= 1 ? 'Y' : undefined);
```
Comment (farisashai) [inline src/pages/store/item/[uuid].tsx]: the variant should default to the first size option so don't have to deal with layout shift of everything moving down when the price becomes visible after selection

Comment (farisashai) [reply src/pages/store/item/[uuid].tsx]: seeking to learn: what is this default value "Y"

src/pages/store/item/[uuid].tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [inCart, setInCart] = useState<boolean>(false);
+  const [amount, setAmount] = useState<number>(1);
```
Comment (farisashai) [inline src/pages/store/item/[uuid].tsx:29]: seeking to learn: why is this state variable defined in this component if the only place it's ever used is in `<AddCartButton>`

Comment (farisashai) [reply src/pages/store/item/[uuid].tsx:29]: same question for the inCart variable

farisashai (changes requested):
great progress so far - left a couple of nitpicks but two major comments to finalize this feature page

1. I don't think the add to cart button should switch the text to "Remove Cart" after being clicked because this introduces a lot more complicated functionality instead of a single purpose for the button. the remove from cart functionality will be handled on the actual cart page. the button action should always add the selected quantity of the variant to the cart and (in a future PR) should eventually open a modal confirming the item is added to the cart with the two options to "View Cart" or "Continue Shopping" (return to store home page) which match the implemented design
2. you need to find a reasonable max width for the page container and set it so the space between the content doesn't grow unbound on wider screens (you can zoom out a lot to test this) .coverContainer cannot have full width for this, you need to have the container hug the image's square aspect ratio

src/components/store/AddCartButton/style.module.scss
```diff
@@ -44,8 +49,8 @@
     border-radius: 10px;
     font-size: vars.$input-text-medium;
     font-weight: bold;
-    height: 40px;
-    width: 200px;
+    padding: 0.3rem 1rem;
+    word-break: break-all;
```
SheepTester: it's better to use

```suggestion
    word-break: break-word;
```

`break-all` ignores words and breaks wherever it wants, even when it's not necessary to break a word (fine for Chinese, not great for English), while `break-word` only breaks words if it's absolutely necessary

[here's a demo/documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break)

SheepTester:
feedback from @farisashai in #124 that better applies to your PR:

> i'm going to leave some comments here that are only relevant for the store item page too and review it all at once
> 
> * when you enter an item page with variants, the first option that isn't out of stock should be selected by default to prevent layout shift from the price appearing after selecting an option

> might be nice to list which collection this item is a part of on the item page if that data exists easily

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  } else {
+    buyButtonText = 'Out of Stock';
+  }
+
+  const optionArr: Array<{ value: string; label: string }> = [];
+
+  if (maxCanBuy === 0) {
+    optionArr.push({ value: `0`, label: `0` });
+  }
+
```
SheepTester: This line isn't needed anymore because the dropdown gets hidden when `maxCanBuy` is 0, right?

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    <div className={styles.addCartGroup}>
+      {maxCanBuy > 0 ? (
+        <p>You can buy up to {maxCanBuy} of this item.</p>
+      ) : (
+        <p>You can&apos;t buy any more of this item!.</p>
+      )}
+      {currSize === undefined ? <p className={styles.error}>Please select a size.</p> : null}
```
SheepTester: You should use the variant type thingy instead of "size." For example, for [apple](https://membership-portal-ui-v2-git-andy-storeitempage-acmucsd.vercel.app/store/item/66a2a2e6-464d-4d74-a9f1-08b8ffa4fc8c) it should say "color"

image

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  inStock: boolean;
+  inCart: boolean;
+  lifetimeRemaining: number;
+  monthlyRemaining: number;
+  amountToBuy: number;
+  onCartChange: (inCart: boolean) => void;
+  onAmountChange: (amountToBuy: number) => void;
+}
+
+const AddCartButton = ({
```
SheepTester: I think you could look into a better name for this component since it now has a bit more than just the add to cart button

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+      {currSize === undefined ? <p className={styles.error}>Please select a size.</p> : null}
+
+      {currSize === undefined ? null : (
+        <div className={styles.buttonRow}>
+          {!inStock || maxCanBuy <= 1 ? null : (
+            <div className={styles.quantityColumn}>
+              <h4>Quantity</h4>
+              <Dropdown
+                name={`options${myID}`}
+                ariaLabel={`Dropdown to select the number of items to purchase for ${myID}`}
```
SheepTester: I think you should use the item name and variant rather than `myID`

src/components/store/AddCartButton/index.tsx
SheepTester: why is this resolved, the `title` attribute is still here

src/components/store/AddCartButton/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+            </div>
+          )}
+
+          <button
+            className={`${inStock && maxCanBuy > 0 ? styles.buttonInStock : styles.buttonNoStock} ${
+              styles.button
+            }`}
+            type="button"
+            title={`${buyButtonText} Button`}
+            value={buyButtonText}
```
SheepTester: I don't think you need to set `value` for a `<button>` here, so you should remove it (as well as `title`)

src/components/store/AddCartButton/style.module.scss
```diff
@@ -0,0 +1,70 @@
...
+      gap: 0.75rem;
+    }
+
+    div > select {
+      font-size: 1rem;
+      font-size: 1.25rem;
+      font-style: normal;
+      font-weight: 700;
+      line-height: 1.25rem;
+      line-height: 200%; /* 2.5rem */
```
SheepTester: Is there a reason why you have multiple `font-size`s and `line-height`s here?

```suggestion
      font-size: 1.25rem;
      font-style: normal;
      font-weight: 700;
      line-height: 2.5rem;
```

src/components/store/AddCartButton/style.module.scss
```diff
@@ -0,0 +1,70 @@
...
+    background-color: var(--theme-primary-2);
+    color: var(--theme-text-on-background1);
+  }
+
+  .buttonNoStock {
+    background-color: var(--theme-background);
+    border: solid;
+    border-color: var(--theme-accent-line-2);
+
+    border-width: 1px;
```
SheepTester: You could use the shorthand here

```suggestion
    border: 1px solid var(--theme-accent-line-2);
```

src/components/store/ItemHeader/style.module.scss
```diff
@@ -0,0 +1,42 @@
...
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+  gap: 1rem;
+
+  h1 {
+    @media (max-width: vars.$breakpoint-md) {
+      font-size: vars.$card-heading-1;
+    }
+    display: inline;
```
SheepTester: What is this for? I think you can remove `display: inline`

src/components/store/SizeSelector/style.module.scss
```diff
@@ -0,0 +1,61 @@
...
+      height: $indicator-size;
+      justify-content: center;
+      min-width: #{$indicator-size * 2};
+      padding: 0 1rem;
+    }
+
+    input {
+      opacity: 0;
+      position: absolute;
+    }
```
SheepTester: You should also put `pointer-events: none;` or something because I'm still able to click on the radio for B even though it's on top of A

image

src/styles/pages/StoreItemPage.module.scss
```diff
@@ -0,0 +1,81 @@
+@use 'src/styles/vars.scss' as vars;
+
+.optionsContainer {
+  display: flex;
+  flex: none;
+  flex-direction: column;
+  gap: 1.1rem;
+  max-width: 22.5rem;
+  min-width: 18rem;
```
SheepTester, line 9: it still feels like the layout depends on the length of the text

image

image

adding `width: 100%` here fixes this

SheepTester (approved):
looks good, just a bunch of nit picks

src/pages/store/item/[uuid].tsx
```diff
@@ -0,0 +1,103 @@
...
+    item.options.length <= 1 ? { type: 'Y', value: 'Y', position: 0 } : undefined
+  );
+  const [inCart, setInCart] = useState<boolean>(false);
+  const [amount, setAmount] = useState<number>(1);
+
+  const currItemOption: PublicMerchItemOption | null | undefined =
+    item.options.find(val => val.metadata?.value === selectedOption?.value) ??
+    item.options.find(option => option.quantity > 0) ??
+    item.options[0] ??
+    null;
```
Comment (farisashai) [inline src/pages/store/item/[uuid].tsx]: wondering what the difference is between selectedOption and currItemOption - i'm not sure if two separate variables are needed here. when I suggested the change to use the first item option to set the price field on page load, I was thinking it would be implemented by setting the initial value of selectedOption to `item.options[0]` instead of undefined on line 26

src/components/store/SizeSelector/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+          onClick={() => onSizeChange(val.metadata)}
+        />
+        <label htmlFor={tempId}>{val.metadata?.value}</label>
+      </Fragment>
+    );
+  });
+
+  return (
+    <div className={styles.sizeSelector}>
+      {options.length > 1 ? (
```
farisashai: if we have access to the options variable in the parent component rendering the `SizeSelector`, can we do this check at that top level instead and say don't render the SizeSelector if this item doesn't support options so we don't have to render an empty div and save it's state

src/components/store/SizeSelector/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { Fragment, useId } from 'react';
+
+interface SizeSelectorProps {
+  currOption: MerchItemOptionMetadata | undefined;
+  options: PublicMerchItemOption[];
+  onOptionChange: (currSize: PublicMerchItemOption) => void;
+}
+function toTitleCase(str: string) {
+  return str.toLowerCase().replace(/\.\s*([a-z])|^[a-z]/gm, s => s.toUpperCase());
+}
```
alexzhang1618: For cleanliness, let's move this to utils and add a comment describing it

src/pages/store/item/[uuid].tsx
```diff
@@ -0,0 +1,102 @@
...
+    item.options.find(option => option.quantity > 0) ?? item.options[0] ?? null
+  );
+  const [inCart, setInCart] = useState<boolean>(false);
+  const [amount, setAmount] = useState<number>(1);
+
+  //   const currItemOption: PublicMerchItemOption | null | undefined =
+  // item.options.find(val => val.metadata?.value === selectedOption?.value) ??
+  // item.options.find(option => option.quantity > 0) ??
+  // item.options[0] ??
+  // null;
```
Comment (alexzhang1618) [inline src/pages/store/item/[uuid].tsx]: Let's clean up these comments if we're not using them

alexzhang1618 (commented):
For variantless items, should this say Add to Cart instead of Select a Size? Or are we waiting for cart implementation to come through first?
Screenshot 2024-02-15 at 1 22 40 PM

Besides this, everything else LGTM

src/components/store/CartOptionsGroup/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  onCartChange,
+  lifetimeRemaining,
+  monthlyRemaining,
+  amountToBuy,
+  optionsKey,
+  onAmountChange,
+}: CartOptionGroupProps) => {
+  const myID = useId();
+
+  const maxCanBuy = Math.min(20, Math.min(lifetimeRemaining, monthlyRemaining));
```
alexzhang1618, line 31: Let's add a comment explaining the business logic for this?

src/components/store/CartOptionsGroup/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+    buyButtonText = 'Add to Cart';
+  } else {
+    buyButtonText = 'Out of Stock';
+  }
+
+  const optionArr: Array<{ value: string; label: string }> = [];
+
+  for (let i = 1; i <= maxCanBuy; i += 1) {
+    optionArr.push({ value: `${i}`, label: `${i}` });
+  }
```
alexzhang1618: nit: I think 
```tsx
Array(maxCanBuy).fill().map((_, i) => i+1);
```
would be more efficient? Can add a comment explaining what this is doing also

alexzhang1618 (approved):
brilliant 🤩 good to merge after these two nits

farisashai:
This should not be a dropdown tbh because there's no reason to list every number when it can get arbitrarily large- it should be a numeric input field that provides an error message if the input is invalid. We should open a separate issue for this fix before merging this since it's not a quick change

image

farisashai:
From the perspective of not over complicating components and isolating a single purpose to each one, I think it would make a lot more sense for the Add to Cart button to never change its text. instead, there should be a Boolean prop passed in whether to disable (grey out) the action or not. The error should be explained on the page itself in the spot where it says "You can buy up to 1 of this item" because right now that text is misleading saying you can buy 1 but it's actually out of stock

image

farisashai:
For whatever reason, it seems like a lot of images that have been uploaded to the store historically have a transparent region in the png. We should account for this by adjusting the design in some way with a border or shadow to highlight the image in separation from the page

image

src/styles/pages/StoreItemPage.module.scss
```diff
@@ -0,0 +1,74 @@
...
+
+.navbarBodyDiv {
+  display: flex;
+
+  flex-direction: column;
+  gap: 2rem;
+  justify-content: center;
+
+  margin-left: auto;
+  margin-right: auto;
```
farisashai: shorthand for this is `margin: 0 auto;`

src/components/store/CartOptionsGroup/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+    buyButtonText = 'Add to Cart';
+  } else {
+    buyButtonText = 'Out of Stock';
+  }
+
+  // Fills values of the Dropdown to be increasing sequential numbers
+  const optionArr: Array<{ value: string; label: string }> = Array(maxCanBuy)
+    .fill({ value: `${1}`, label: `${1}` })
+    .map((i, index) => {
+      return { value: `${Number(i.value) + index}`, label: `${Number(i.label) + index}` };
```
farisashai: I don't think you need the fill here - you can just take the original array and map every element to index + 1 since that's already hardcoded anyways. Don't need to store it in a variable

src/pages/store/item/[uuid].tsx
```diff
@@ -0,0 +1,108 @@
...
+import { useState } from 'react';
+
+interface ItemPageProps {
+  user: PrivateProfile;
+  item: PublicMerchItemWithPurchaseLimits;
+}
+
+const StoreItemPage = ({ user: { credits }, item }: ItemPageProps) => {
+  const [selectedOption, setSelectedOption] = useState<PublicMerchItemOption | null | undefined>(
+    item.options.find(option => option.quantity > 0) ?? item.options[0] ?? null
```
Comment (farisashai) [inline src/pages/store/item/[uuid].tsx:26]: just wondering, when would this default to a value that isn't just the first element in the array ?

src/components/store/CartOptionsGroup/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+                ariaLabel={`Dropdown to select the number of items to purchase for ${myID}`}
+                options={optionArr}
+                value={`${amountToBuy}`}
+                onChange={val => onAmountChange(Number(val))}
+              />
+            </div>
+          )}
+
+          <button
+            className={`${inStock && maxCanBuy > 0 ? styles.buttonInStock : styles.buttonNoStock} ${
```
farisashai: you used this Boolean condition (or its negation) at least three times in this file from what I've seen. you should probably store it in a Boolean variable isPurchasable and reference that everywhere instead to keep the logic consistent as to exactly what it's checking

farisashai (commented):
some quick tiny comments but everything seems good now - will approve after we have issues covering all of the earlier comments I mentioned

farisashai:
I think we're all aware of this last point but we also need an issue for the add to cart button to open a modal that lets the user confirm adding the item to the actual cart, however Raymond decides to implement that

src/pages/store/item/[uuid].tsx
Comment (SheepTester) [reply src/pages/store/item/[uuid].tsx:26]: when `item.options` is empty

our typescript rules make `item.options[0]` a `PublicMerchItemOption | undefined`. but also, backend doesn't enforce that items must have at least one option, so there can be [items with no options](https://membership-portal-ui-v2-git-andy-storeitempage-acmucsd.vercel.app/store/item/13b32fdc-a220-48e4-b432-88aa33019e22)

farisashai (approved):
LGTM 🚀

PR #96, opened 2023-12-20
Create EventCard Component

alexzhang1618:
# Changes

- Fixed some sizing bugs in Typography (small was larger than large, ...)
- Added CommunityLogo in components to render the logo for a community
- Added an event display on the dashboard along with event cards that pop up modals when clicked
  - Components for this are under components/event
- Added the `ics` package to support calendar exports for events
- Add a new API call to get attendances for the current user

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots
Screenshot 2023-12-20 at 3 45 41 PM
Screenshot 2023-12-20 at 3 46 00 PM
Screenshot 2023-12-20 at 3 46 12 PM


src/components/events/CalendarButtons/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+        className={styles.calendarButton}
+        onClick={() => saveToGoogleCal(event)}
+      >
+        <GoogleCalendarLogo alt="google calendar" />
+        <b>Add to Google Calendar</b>
+      </button>
+      <button type="button" className={styles.calendarButton} onClick={() => saveToAppleCal(event)}>
+        <AppleCalendarLogo alt="apple calendar" />
+        <b>Add to Apple Calendar</b>
+      </button>
```
SheepTester: Should these be links? This way, users can see that they'll be sent to Google Calendar, or they can copy the URL or share the link to an app or another device

The "Add to Apple Calendar" button could also be a download link as well since `createDownloadFile` also creates an `<a>` element

src/components/events/CalendarButtons/style.module.scss
```diff
@@ -0,0 +1,35 @@
...
+}
+
+.calendarButton {
+  align-items: center;
+  background: transparent;
+  border: 1px solid var(--theme-primary-2);
+  border-radius: 10px;
+  display: flex;
+  gap: 10px;
+  justify-content: center;
```
SheepTester: Maybe it should be left aligned so the icons and "Add to" lies up?

```suggestion
  justify-content: flex-start;
```

though for some reason the icons don't seem to be properly aligned (the Google logo is off center)

image

src/components/events/EventCard/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+const EventCard = ({ event, attended }: EventCardProps) => {
+  const { cover, title, start, end, location } = event;
+  const [expanded, setExpanded] = useState(false);
+
+  let displayCover = cover;
+  if (!displayCover) {
+    displayCover = '/assets/graphics/store/hero-photo.jpg';
+  }
```
SheepTester: this could be simplified to
```suggestion
  let displayCover = cover || '/assets/graphics/store/hero-photo.jpg';
```

src/components/events/EventCard/style.module.scss
```diff
@@ -0,0 +1,52 @@
...
+        white-space: nowrap;
+      }
+    }
+  }
+
+  .viewMore {
+    background: none;
+    display: flex;
+    padding: 0;
+  }
```
SheepTester: what is this used for

src/components/events/EventCard/style.module.scss
```diff
@@ -0,0 +1,52 @@
+.container {
+  background: transparent;
+  border: 1px solid var(--theme-primary-6);
+  border-radius: 10px;
+  -ms-overflow-style: none;
+  overflow-x: hidden;
+  padding: 0;
+  position: relative;
+  scrollbar-width: none;
```
SheepTester: Is there a reason why a scrollbar would appear?

src/components/events/EventCard/style.module.scss
```diff
@@ -0,0 +1,52 @@
...
+
+    .header {
+      align-items: center;
+      display: flex;
+      flex-direction: row;
+      gap: 0.5rem;
+
+      .eventDetails {
+        overflow: hidden;
+        text-overflow: ellipsis;
```
SheepTester: The ellipses don't show for me

image

It seems you have to set these two CSS properties directly on the `Typography` elements themselves in order for it to work

src/components/events/CalendarButtons/style.module.scss
```diff
@@ -0,0 +1,35 @@
...
+  padding: 0.75rem;
+  width: 15rem;
+}
+
+.calendarButton:hover {
+  background: var(--theme-primary-2);
+  transition: all 0.2s;
```
SheepTester: you should move this `transition` property to `.calendarButton` itself. this way, there's also an animation when your mouse moves off the button

Also, you could move this rule inside `.calendarButton` like this:

```scss
.calendarButton {
  ...
  transition: all 0.2s;
  
  &:hover {
    background: var(--theme-primary-2);
  }
}
```

This has the same result, but it makes it more clear that these two rules are related

src/components/events/EventCard/style.module.scss
```diff
@@ -0,0 +1,52 @@
...
+    background: none;
+    display: flex;
+    padding: 0;
+  }
+}
+
+.container:hover {
+  box-shadow: 0 0 8px var(--theme-primary-6);
+  transform: scale(1.025);
+  transition: transform 0.3s;
```
SheepTester: you should also move this `transition` property to `.container` so there's an animation when you move your mouse off the event card

src/components/events/EventModal/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  let displayCover = cover;
+  if (!displayCover) {
+    displayCover = '/assets/graphics/store/hero-photo.jpg';
+  }
+
+  let displayEventLink = eventLink;
+  if (!eventLink) {
+    displayEventLink = `https://acmucsd.com/events/${event.uuid}`;
+  }
```
SheepTester:
```suggestion
  let displayCover = cover || '/assets/graphics/store/hero-photo.jpg';

  let displayEventLink = eventLink || `https://acmucsd.com/events/${event.uuid}`;
```

you could also use `??` instead of `||` if you're feeling fancy

src/components/events/EventModal/index.tsx
SheepTester, line 1: In #91 I'm adding a generic `Modal` component so there's less code duplication between this and `HelpModal`. I can migrate this component to use `Modal` if this PR gets merged first

src/components/events/EventModal/style.module.scss
```diff
@@ -0,0 +1,108 @@
...
+      position: relative;
+    }
+
+    .contents {
+      display: flex;
+      flex-direction: column;
+      gap: 2.5rem;
+      max-width: 70rem;
+      padding: 2.5rem;
+      width: 70vw;
```
SheepTester, line 63: the modal is very narrow on mobile. this is because of this `width: 70vw` line, but you could also remove the horizontal padding on `.modal` when the screen is narrow enough, so there's more horizontal screen space to work with

current | suggested
--- | ---
image | image

src/components/events/EventModal/style.module.scss
```diff
@@ -0,0 +1,108 @@
...
+
+      .header {
```
SheepTester, line 65: image

you should add some gap between the title and the calendar buttons

src/components/events/PointsDisplay/index.tsx
```diff
@@ -0,0 +1,17 @@
+import styles from './style.module.scss';
+
+interface PointsDisplayProps {
+  points: number;
+  attended: boolean;
+}
+
+const PointsDisplay = ({ points, attended }: PointsDisplayProps) => {
+  return (
+    <div className={`${styles.circle} ${attended ? styles.green : styles.blue}`}>
```
SheepTester: I'd explore alternate ways to show whether you've attended an event (unless we want to keep it hidden?). Green vs blue isn't exactly clear, and relying on color isn't great for colorblind people. One idea is to show a check icon if you've attended an event

src/components/events/EventModal/index.tsx
SheepTester: Should this also use `formatURLEventTitle` to include the event title in the URL, like what is used for the Add to Apple Calendar button?

SheepTester (changes requested):

src/components/common/CommunityLogo/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import AILogo from '@/public/assets/acm-logos/communities/ai.png';
+import CyberLogo from '@/public/assets/acm-logos/communities/cyber.png';
+import DesignLogo from '@/public/assets/acm-logos/communities/design.png';
+import HackLogo from '@/public/assets/acm-logos/communities/hack.png';
+import ACMLogo from '@/public/assets/acm-logos/general/light-mode.png';
+
+interface CommunityLogoProps {
+  community: string;
+  width: number;
```
farisashai: nit: I would rename this to size if it's going to define both the width and height as the same value

src/components/events/EventModal/index.tsx
```diff
@@ -0,0 +1,108 @@
+import { Typography } from '@/components/common';
+import CommunityLogo from '@/components/common/CommunityLogo';
```
farisashai:
```suggestion
import { CommunityLogo } from '@/components/common';
```
super tiny nit: try to import from the enclosing folder to ensure imports are always named correctly

src/components/events/EventModal/style.module.scss
```diff
@@ -0,0 +1,108 @@
...
+        display: flex;
+        flex-direction: row;
+        gap: 0.5rem;
+        text-decoration: underline;
+      }
+    }
+  }
+}
+
+@media screen and (max-width: 60rem) {
```
farisashai: nit: don't hardcode breakpoints like this otherwise things get really messy really fast - there are preset breakpoints that are standard conventions defined in the global variables vars.scss

$breakpoint-md: 768px; // Use this breakpoint for most mobile-first CSS designs

src/components/events/EventModal/style.module.scss
```diff
@@ -0,0 +1,108 @@
...
+  overscroll-behavior: contain;
+  padding: 2rem;
+  pointer-events: none;
+  position: fixed;
+  top: 0;
+  transition: all 0.2s;
+  user-select: none;
+  visibility: hidden;
+  width: 100vw;
+  z-index: 1000; // Just one above the header's 999
```
farisashai, line 19: this seems like the start of a very difficult z-index system to maintain - possibly consider any other approaches

src/components/events/PointsDisplay/style.module.scss
```diff
@@ -0,0 +1,33 @@
...
+  justify-content: center;
+  position: absolute;
+  right: 2.5%;
+  top: 5%;
+  width: 50px;
+  z-index: 1;
+
+  &.blue {
+    background-color: var(--theme-primary-4);
+  }
```
farisashai:
```suggestion
  background-color: var(--theme-primary-4);
```


you don't need a class name for blue since this is essentially the default background color for all .circle elements - i would only keep the green one as a more specific override but also rename it for specificity to attended or something better. i agree w sean that we shouldn't be using colors to indicate attendance anyways

src/components/events/EventModal/style.module.scss
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+}
+
+@media screen and (max-width: 60rem) {
+  .eventDetails {
+    flex-wrap: wrap;
+
+    .eventTitle {
+      font-size: 1.5rem !important;
```
farisashai, line 96: what's the reason for the !important? it's always never necessary if you specific class names and elements correctly so i would try to remove it and solve this differently

farisashai (changes requested):
sean covered everything visual mostly - biggest feedback is reuse his modal component or find a way to unify the two and resolve the scrolling underneath the modal bug

src/components/events/CalendarButtons/index.tsx
alexzhang1618: 👍

src/components/events/CalendarButtons/style.module.scss
alexzhang1618: I've updated the logos since there was a weird gap in the Google logo, and fixed the alignment

src/components/events/EventCard/style.module.scss
alexzhang1618: I think it's leftover CSS, deleted 😓

src/components/events/EventCard/style.module.scss
alexzhang1618: removed 👍

src/components/events/EventCard/style.module.scss
alexzhang1618: fixed o7

src/components/events/CalendarButtons/style.module.scss
alexzhang1618: good catch! fixed

src/components/events/EventModal/index.tsx
alexzhang1618: hm `displayEventLink` just links to a main website preview of the event, so we need to keep it this way by convention, so probably not

src/components/events/EventModal/style.module.scss
alexzhang1618: fixed !

src/components/events/EventModal/index.tsx
alexzhang1618: Sure, let's revisit this after this PR and #91 are merged and done

src/components/events/EventModal/style.module.scss
alexzhang1618: done!

src/components/events/PointsDisplay/index.tsx
alexzhang1618: I think we should iterate a little with design before settling here, I like the idea of a check but it does make the point total unreadable 🤔

alexzhang1618: Maybe for now we can keep green and update in a future PR once we've figured out a direction on this

src/components/events/EventModal/style.module.scss
alexzhang1618: good catch, I didn't know this existed 😮

src/components/events/EventModal/style.module.scss
alexzhang1618: So for the `<Typography>` component, it specifies its own styles for the text, so I need to add `!important` when trying to downsize for mobile views, or else it won't update.

I could try to update the variant for `Typography` instead of using CSS here, but I think this is a little simpler. I'll add some comments here also.

src/components/events/EventModal/style.module.scss
alexzhang1618: @SheepTester maybe some food for thought 😁

farisashai (approved):
looks fine at a quick glance, approving so it's not blocked - sean can do a sanity check if he wants

src/components/events/EventModal/style.module.scss
SheepTester: we could store z-indices as constants in vars.scss

src/components/events/CalendarButtons/index.tsx
```diff
@@ -54,31 +52,45 @@ const saveToAppleCal = (event: PublicEvent): void => {
...
   const response = ics.createEvent(attributes);
   if (response.value) {
-    createDownloadFile(response.value, formatURLEventTitle(event.title));
-  } else {
-    showToast(response.error?.message || 'An error occurred when saving to Apple Calendar.');
+    // Referencing this answer to save text in a file and download it
+    // https://stackoverflow.com/questions/44656610/download-a-string-as-txt-file-in-react
+    const file = new Blob([response.value], { type: 'text/calendar' });
+    const href = URL.createObjectURL(file);
+    const download = `${formatURLEventTitle(event.title)}.ics`;
```
SheepTester, line 65: you should probably put this in a `useMemo` or something. it's probably not a big deal because the ics file probably isn't that big, and there aren't that many interactive elements on the page, but every time React re-renders, it will create a new object url, which would lead to a memory leak

Something like

```ts
const url = useMemo(() => {
  const file = ...;
  return URL.createObjectURL(file);
}, [...]);

useEffect(() => {
  return () => {
    URL.revokeObjectURL(url);
  };
}, [url]);
```

maybe

SheepTester (approved):
lgtm

src/components/events/CalendarButtons/index.tsx
alexzhang1618: I've been playing around with this for like an hour and for some reason this approach isn't really working.

It seems like the blob/objectURL no longer works after refreshing the page, so after hitting refresh none of the apple calendar buttons work.

I'm unsure if this is a local-only problem because the console error after hitting the apple calendar link after a refresh is `Not allowed to load local resource: blob:nodedata:...`

I'm going to push what I have sans the `useMemo` for appleCal and we can revisit in the future

PR #101, opened 2023-12-29
Add events carousel layout to home page

alexzhang1618:
# Info

# Description

motivation or context.

- Implemented events display on the store homepage
- Added a separate Events page to make a big, searchable archive for all events (in a future PR)
- Add Dashboard as a navbar item (should it be named Home?)
- Add a common Carousel component to be used both in the store and portal homepages
- Migrated Store Collection Slider to use Carousel

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots
Screenshot 2023-12-29 at 2 41 00 PM
Screenshot 2023-12-29 at 2 41 09 PM


src/components/layout/Navbar/index.tsx
```diff
@@ -77,7 +78,9 @@ const Navbar = ({ user }: NavbarProps) => {
         </Link>
         {/* Desktop Nav Links */}
         <nav className={styles.portalLinks}>
-          <Link href={config.homeRoute}>Events</Link>
+          <Link href={config.homeRoute}>Dashboard</Link>
+          <p aria-hidden>·</p>
+          <Link href={config.eventsRoute}>Events</Link>
           <p aria-hidden>·</p>
           <Link href="/leaderboard">Leaderboard</Link>
           <span aria-hidden>·</span>
```
SheepTester: this isn't part of the PR, but it's weird that we use a mix of `p` and `span` for these dot separators

src/components/layout/Navbar/index.tsx
```diff
@@ -104,6 +107,14 @@ const Navbar = ({ user }: NavbarProps) => {
           className={styles.mobileNavItem}
           onClick={() => setMenuOpen(false)}
           href={config.homeRoute}
+        >
+          <DashboardIcon className={styles.iconLink} />
+          Dashboard
+        </Link>
+        <Link
+          className={styles.mobileNavItem}
+          onClick={() => setMenuOpen(false)}
```
SheepTester: We could probably refactor this part of the navbar by taking advantage of event propagation and moving the repeated `onClick={() => setMenuOpen(false)}` to the parent `<nav>`. If we only want the mobile navbar to close when the user presses a link, you could add a check for that using [Element.closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)

```tsx
<nav className={styles.portalLinks} onClick={e => e.target.closest('a') && setMenuOpen(false)}>
```

I don't think this will type check, though, since using `event.target` in TypeScript is kind of hairy. might have to do something like `e.target instanceof Element && ...`

SheepTester (approved):
The code looks fine to me. I'm assuming the dashboard design is still somewhat temporary since I don't think we have a design for it yet from design team.

I think it might be better to call the dashboard "Home" because users might be looking for a way back to the landing page, and calling it anything other than "Home" would lead them lost

Just some refactoring suggestions for the navbar 👇

src/components/layout/Navbar/index.tsx
alexzhang1618: done! I disabled an ESLint rule here since they were complaining about `onClick`s for a div and I didn't want to get too deep into the weeds on that lol

src/components/common/Carousel/index.tsx
```diff
@@ -0,0 +1,12 @@
+import { ReactNode } from 'react';
```
farisashai:
```suggestion
import type { ReactNode } from 'react';
```

src/components/events/EventCarousel/index.tsx
```diff
@@ -0,0 +1,41 @@
+import { Carousel, Typography } from '@/components/common';
+import EventCard from '@/components/events/EventCard';
```
farisashai:
```suggestion
import { EventCard } from '@/components/events';
```

src/components/events/EventCarousel/style.module.scss
```diff
@@ -0,0 +1,37 @@
...
+      padding: 0;
+
+      &:hover {
+        text-decoration: underline;
+      }
+    }
+  }
+}
+
+.card {
```
farisashai: this class name should be nested under .wrapper within the structure of how the component is being used

src/pages/index.tsx
```diff
@@ -1,34 +1,60 @@
...
   attendances,
 }: HomePageProps) => {
   return (
-    <div>
-      <h1>Portal Home Page</h1>
-      <pre>User Info: {JSON.stringify(user, null, 2)}</pre>
+    <div className={styles.page}>
+      <div>
+        <h1>Portal Home Page</h1>
+        <pre>User Info: {JSON.stringify(user, null, 2)}</pre>
```
farisashai:
```suggestion
```
Delete this temp line now that you're putting more real content on the page

src/pages/index.tsx
```diff
@@ -1,34 +1,60 @@
...
+          events={liveEvents}
+          attendances={attendances}
+        />
+      )}
+
+      {upcomingEvents.length > 0 && (
+        <EventCarousel
+          title="Upcoming Events"
+          description="Mark your calendars! These events are just around the corner!"
+          events={upcomingEvents}
```
farisashai: even though we only expect a few incoming events, I would still slice this to cap at 10 for consistency

src/styles/pages/index.module.scss
farisashai, line 1: Rename this file to Home.module.scss

farisashai (changes requested):
Some nits and design decision thoughts

I like the idea of a permanent indexable events page and using the home page as a dashboard that can be updated to convey useful info to members. 

- Can you fix the blue text color in the event cards
image
- I personally don't see a reason to show past events on the home page of we have a full events page. I feel like it's simpler if the home dashboard can just help them plan for attending events. Also, tracking attendances for upcoming events doesn't make much sense but maybe that's not too significant right now
- I think the live event section can have a more engaging view as a banner later on kinda since we'll never have more than 2 (usually 1) live events so a scroll carousel doesn't help with the UX much.

src/components/events/EventCard/index.tsx
```diff
@@ -26,7 +27,11 @@ const EventCard = ({ event, attended }: EventCardProps) => {
         event={event}
         onClose={() => setExpanded(false)}
       />
-      <button type="button" className={styles.container} onClick={() => setExpanded(true)}>
+      <button
+        type="button"
+        className={`${styles.container} ${className}`}
```
farisashai: Can you check if this returns an undefined className when you inspect the preview? It should be an empty string if the className isn't provided

alexzhang1618: It does say undefined, so I added a little fix in EventCard

farisashai (approved):
lgtm! 🤑

PR #112, opened 2024-01-02
Finish main page hero

alexzhang1618:
# Description

motivation or context.

- Implement the main page hero
- This also includes check-in handling through the QR code/link, and form on the main page

Things to do:
- cool and funny animations on the hero

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots

Normal screens
Screenshot 2024-01-02 at 12 24 48 PM
Wide screens
Screenshot 2024-01-02 at 12 25 07 PM
Mobile screens
Screenshot 2024-01-02 at 12 25 28 PM


farisashai (commented):
I really like how you linked every other page from the home page in a good way without needing to use the navbar. I honestly think we should have a "about acm" or "learn more" link in this home section and completely remove the link from the main navigation since it's so insignificant

src/pages/index.tsx
```diff
@@ -1,22 +1,73 @@
...
   pastEvents: PublicEvent[];
   upcomingEvents: PublicEvent[];
   liveEvents: PublicEvent[];
   attendances: PublicAttendance[];
 }
 
-const PortalHomePage = ({ pastEvents, upcomingEvents, liveEvents, attendances }: HomePageProps) => {
+const checkin = async (token: string, code: string) => {
+  try {
+    const response = await attendEvent(token, code);
```
farisashai: Are we still following the convention of using the EventManager files to wrap api calls or scrapping that idea completely? I think it is a cleaner structure to isolate logic for the api arguments and error handling outside of page files so we can focus on the outputs only.

farisashai (changes requested):

src/components/home/Hero/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+        priority
+        fill
+      />
+      <Image
+        className={`${styles.image} ${styles.mobile}`}
+        src="/assets/graphics/portal/waves.svg"
+        alt="Landing page graphic"
+        priority
+        fill
+      />
```
SheepTester: you could also remove the waves from the raccoon image so you can keep waves.svg visible always, and it allows for the raccoon to move more independently if we want to

src/components/home/Hero/style.module.scss
```diff
@@ -0,0 +1,154 @@
...
+        box-shadow: 0 2px 5px rgba(0, 0, 0, 0.3);
+        display: flex;
+        flex-direction: column;
+        gap: 0.75rem;
+        padding: 2rem;
+        transition: box-shadow 0.2s;
+
+        &:hover {
+          box-shadow: 0 5px 10px rgba(0, 0, 0, 0.3);
+        }
```
SheepTester: it's a bit weird having a hover effect for the check-in card since clicking it doesn't do anything

src/components/home/Hero/style.module.scss
```diff
@@ -0,0 +1,154 @@
...
+      display: flex;
+      flex-direction: column;
+      gap: 1rem;
+      max-width: fit-content;
+      padding-bottom: 2rem;
+
+      .checkin {
+        background-color: var(--theme-accent-line-2-transparent);
+        border-radius: 1rem;
+        box-shadow: 0 2px 5px rgba(0, 0, 0, 0.3);
```
SheepTester: we have `var(--theme-shadow)` which you could use for shadows

src/components/home/Hero/style.module.scss
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  .image {
+    margin: 0 auto;
+    max-height: 1080px;
+    max-width: 1920px;
+    object-fit: cover;
+    object-position: 100% 100%;
+    // We have to manually add width since it's taking the width of the container, which is padded.
+    width: calc(100% + 4rem) !important;
```
SheepTester: you could avoid using `!important` here by instead applying negative margin on `.hero`:

```css
.hero {
  margin: 0 -4rem;
  padding: 0 4rem;
}
```

src/components/home/Hero/style.module.scss
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+}
+
+@media screen and (width <= vars.$breakpoint-sm) {
+  .actions {
+    .checkin {
+      .checkinButtons {
+        .submit {
+          display: none !important;
```
SheepTester: if you want to avoid using `!important`, you can move these `@media` queries into the styles above, like

```scss
.submit {
  ...
  
  @media screen and (width <= vars.$breakpoint-sm) {
    display: none;
  }
}
```

so you don't have to fight CSS precedence rules. it also keeps the styles for the elements next to each other rather than at the end of the file

src/pages/index.tsx
```diff
@@ -1,22 +1,68 @@
...
+  useEffect(() => {
+    if (typeof router.query.code === 'string' && router.query.code) {
+      // In dev mode, this runs twice because of reactStrictMode in nextConfig.
+      // This will only be run once in prod or deployment.
+      checkin(authToken, router.query.code).then(() => {
+        // Clear out the query params without re-running getServerSideProps.
+        router.replace('/', undefined, { shallow: true });
+      });
+    }
+  }, [authToken, router]);
```
SheepTester: do you think this could be done on the server side, in `getServerSidePropsFunc`?

src/components/home/Hero/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+              {'You have '}
+            </Typography>
+            <Typography variant="h3/bold" className={styles.inline}>
+              {user.points}
+            </Typography>
+            <Typography variant="h3/regular" className={styles.inline}>
+              {' membership points.'}
+            </Typography>
+          </div>
+        </div>
```
SheepTester: is it possible for a user to only have one point? (i.e. is it possible for someone to see "You have 1 membership points"?)

You should use `component="span"` instead of setting `display: inline` on divs. Alternatively, you could reduce code repetition by putting `Link` and `<strong>` inside one `Typography` element, like this:

```suggestion
        <div className={styles.header}>
          <Typography variant="display/light/large" className={styles.headline}>
            Welcome to ACM,&nbsp;
            <Link href={config.profileRoute}>
              <strong>
                {user.firstName}
              </strong>
            </Link>
            !
          </Typography>
          <Typography variant="h3/regular">
            You have&nbsp;
            <strong>
              {user.points}
            </strong>
            &nbsp;membership points.
          </Typography>
        </div>
```

SheepTester (commented):
design suggestions:

image

you could change the padding for `.checkin` to `padding: 1.5rem` so it lines up with the link icons

there's a lot of different font sizes in the hero. I think you should make all the smaller font sizes in the hero the same size for consistency. the smaller the better, because having text too big can feel claustrophobic and ruin the hierarchy (people might skim over it like they're headings or ads, and it can be hard to read)

I'd also reduce the line height of the big "Welcome to ACM" text, as well as increasing the spacing between that and the membership points

Also I think you mixed up the store and leaderboard icons

src/pages/index.tsx
farisashai: I would do the checkin code server side and return an event object that you use to immediately pop up the modal with the successfully checked into X

src/components/home/Hero/index.tsx
alexzhang1618: I'm going to use component="span" since the text wrapping for the other option was a little off

nishantbalaji:
Welcome text / fonts on mobile are really big, pushing more crucial info down into a scroll. Ideally we can reduce size here. Having ACM and the user's name on separate lines looks kinda awkward. 


image

nishantbalaji:
my fault top g

alexzhang1618:
Since the last review:
- Addressed all the visual/quality feedback
- Moved the query param check-in to GSSR
  - clear query params using window.history.replaceState
- Added some code to update the point count and attendance w/o API calls after filling out the check-in form
- Removed About Us from the navbar and instead added a link on the hero

src/components/home/Hero/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+      />
+      <div className={styles.content}>
+        <div className={styles.header}>
+          <Typography variant="display/light/medium" className={styles.heading} component="span">
+            {'Welcome to ACM, '}
+            <Link href={config.profileRoute}>
+              <Typography
+                variant="display/heavy/medium"
+                className={styles.heading}
+                component="span"
```
farisashai: do we have the ability to render a link directly with the typography component (no special styling just pass additional href)? I feel like that would help

src/components/home/Hero/style.module.scss
```diff
@@ -0,0 +1,148 @@
+@use 'src/styles/vars.scss' as vars;
+
+.inline {
+  display: inline;
+}
+
+@media screen and (width <= vars.$breakpoint-md) {
+  .subheading {
+    font-size: 1rem !important;
```
farisashai, line 5: maybe future issue but this needs a better solution

src/pages/index.tsx
```diff
@@ -1,28 +1,94 @@
...
+  checkInResponse,
+}: HomePageProps) => {
+  const [points, setPoints] = useState<number>(user.points);
+  const [attendance, setAttendance] = useState<PublicAttendance[]>(attendances);
+
+  const checkin = async (token: string, attendanceCode: string): Promise<void> => {
+    const response = await attendEvent({ token, attendanceCode });
+    const event = processCheckInResponse(response);
+    if (event) {
+      // This client side code prevents us from having to refetch user and attendance from the API.
```
farisashai: just to confirm you're doing all of these on checkin

- user's membership points are updated live on the page
- users attendance is updated for the event card below
- user cookie is updated so membership points are accurate on all future pages

src/pages/index.tsx
```diff
@@ -49,11 +115,27 @@ const PortalHomePage = ({ pastEvents, upcomingEvents, liveEvents, attendances }:
...
-const getServerSidePropsFunc: GetServerSideProps = async ({ req, res }) => {
+const getServerSidePropsFunc: GetServerSideProps = async ({ req, res, query }) => {
   const authToken = CookieService.getServerCookie(CookieType.ACCESS_TOKEN, { req, res });
-  const events = await EventAPI.getAllEvents();
-  const attendances = await EventAPI.getAttendancesForUser(authToken);
 
+  let checkInResponse = null;
+  if (typeof query.code === 'string') {
+    // If a check-in code is specified, first check in to that event.
+    checkInResponse = await attendEvent({ token: authToken, attendanceCode: query.code });
```
farisashai, line 124: Can you wrap this API call in an async function that returns undefined if the code query parameters isn't given so it can be batched together with the other 3 api calls

src/pages/index.tsx
```diff
@@ -1,28 +1,94 @@
...
+  upcomingEvents,
+  liveEvents,
+  attendances,
+  checkInResponse,
+}: HomePageProps) => {
+  const [points, setPoints] = useState<number>(user.points);
+  const [attendance, setAttendance] = useState<PublicAttendance[]>(attendances);
+
+  const checkin = async (token: string, attendanceCode: string): Promise<void> => {
+    const response = await attendEvent({ token, attendanceCode });
```
farisashai, line 55: the auth token can be fetched from the cookie client side in this function where it's used directly so we don't have to drill a ton of props through the component

src/pages/index.tsx
```diff
@@ -1,28 +1,94 @@
...
+    const event = processCheckInResponse(response);
+    if (event) {
+      // This client side code prevents us from having to refetch user and attendance from the API.
+      setPoints(p => p + event.pointValue);
+      const a: PublicAttendance = {
```
farisashai: rename to something better or inline your usage

src/pages/index.tsx
```diff
@@ -1,28 +1,94 @@
...
+  response: PublicEvent | CustomErrorBody
+): PublicEvent | undefined => {
+  if ('uuid' in response) {
+    // If the response contains a uuid, the response is a PublicEvent.
+    const title = `Checked in to ${response.title}!`;
+    const subtitle = `Thanks for checking in! You earned ${response.pointValue} points.`;
+    showToast(title, subtitle);
+    return response;
+  }
+  showToast('Unable to checkin!', response.message);
```
farisashai, line 38: I thought this code was moved to the EventManager ? This feels a lot more confusing and overcomplicated separated from the actual checkin function and having to check whether we're in a success state or failure state directly in the same function

farisashai (changes requested):

src/pages/index.tsx
alexzhang1618: Since we also do a checkin server side, we can't use showToast in the EventManager if I'm using it in both places

src/pages/index.tsx
alexzhang1618: I can do that, but the intention is to run the check-in API code first since the other API calls (user + attendance) may be different depending on the outcome of the check-in API call

src/pages/index.tsx
alexzhang1618: Membership and attendance are being updated, I'm not updating the cookie since I think we agreed on just pulling from the API to fetch user info like membership points for a page (if we rely on the cookie then across different devices it won't be accurate)

alexzhang1618: This is mainly just a little piece so info is visually accurate right after checking in to an event without having to re-call the API. If we traveled to another page where we needed the info, I'd assume we'd re-fetch

src/components/home/Hero/index.tsx
alexzhang1618: I'll add a little enhancement to Typography so we can do 
```tsx
<Typography variant="display/heavy/medium" className={styles.heading} component="a" href={config.profileRoute} />
```

farisashai (approved):
if the typography component is specified as a "a", we should actually be rendering a `<Link>` component to handle client side routing for us but otherwise changes look good

src/components/home/Hero/style.module.scss
```diff
@@ -0,0 +1,148 @@
+@use 'src/styles/vars.scss' as vars;
+
+.inline {
+  display: inline;
+}
```
SheepTester: don't think this is needed anymore

src/components/home/Hero/style.module.scss
```diff
@@ -0,0 +1,148 @@
...
+      gap: 2rem;
+    }
+
+    .header {
+      display: flex;
+      flex-direction: column;
+      gap: 0.5rem;
+
+      .heading {
+        display: inline;
```
SheepTester: nor is this, because you're using `component="span"` now

src/components/home/Hero/style.module.scss
SheepTester: I think it's one of the downsides of the `Typography` component, that it's hard to define multiple sizes for different screen sizes. Perhaps ideally, `Typography` should use globally defined classes instead of setting `style`, so it can be more easily overridden, and the CSS file can use media queries to define different font sizes for different screen sizes

src/components/common/Typography/index.tsx
```diff
@@ -232,6 +233,21 @@ const variantToCSS = (variant: Variant): CSSProperties => {
 const Typography = (props: PropsWithChildren<TypographyProps>) => {
   const { variant, component, style, children, ...restProps } = props;
 
+  if (restProps.href) {
+    return (
```
farisashai, line 237: why don't you just say Component = Link instead of rewriting the whole thing

alexzhang1618: Typescript complains because href isn't guaranteed to be true and link needs an href

PR #113, opened 2024-01-04
Faris/cleanup optimizations

farisashai:
some cleanup and optimizations for repo

- serversideprops should return as little data as possible, moved event slicing to serverside since the slice range is already hardcoded in
- shallow navigate page routes with native nextjs router to avoid recalculating serversideprops
- default event page to all time so the page isn't empty on initial load ever
- format event links without https:// correctly instead of assuming relative links
- fix hydration error trying to run the browser blob api serverside pre-rendering
- load gifs correctly in top 3 user cards

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -14,7 +14,6 @@ import type {
 import { CookieType } from '@/lib/types/enums';
 import styles from '@/styles/pages/Home.module.scss';
 import { GetServerSideProps } from 'next';
-import { useRouter } from 'next/router';
 import { useEffect, useState } from 'react';
 
 interface HomePageProps {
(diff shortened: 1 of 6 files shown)


src/pages/index.tsx
```diff
@@ -69,15 +70,17 @@ const PortalHomePage = ({
...
+  const router = useRouter();
+
   useEffect(() => {
     if (checkInResponse) {
       // In dev mode, this runs twice because of reactStrictMode in nextConfig.
       // This will only be run once in prod or deployment.
       processCheckInResponse(checkInResponse);
       // Clear the query params without re-triggering getServerSideProps.
-      window.history.replaceState(null, '', '/');
+      router.push(`${config.homeRoute}`, undefined, { shallow: true });
```
alexzhang1618: Does this work? in my testing, after you check-in with the query params and `{ shallow: true }` re-renders w/o reloading props so the toast gets triggered infinitely

farisashai: idk i will remove

src/pages/leaderboard.tsx
```diff
@@ -111,7 +111,7 @@ const LeaderboardPage = ({ sort, leaderboard, user: { uuid } }: LeaderboardProps
           ]}
           value={sort}
           onChange={sort => {
-            router.push(`${config.leaderboardRoute}?sort=${sort}`);
+            router.push(`${config.leaderboardRoute}?sort=${sort}`, undefined, { shallow: true });
```
alexzhang1618: this bricks the leaderboard on the preview

alexzhang1618 (approved):
lgtm 😁

PR #115, opened 2024-01-07
Testing Cypress PR action

farisashai:
Configured Cypress for the project

`yarn test` will run the full suite of Cypress tests in your command-line (the server must be running on `localhost:3000` on the testing API already otherwise the tests will fail)
`yarn test:ui` will open the Cypress UI application to help you easily view the state of every DOM element at every step of your test and make them really easy to write.
`cypress.yml` file is a Github Action used to run the full Cypress suite every time a pull request is opened


raymosun (approved):
lgtm, this should be very useful 🚀

PR #124, opened 2024-01-18
Admin store item and collection management

### LINKED ISSUE #123: Store admin page with multiple store gallery image support
[discord](https://discord.com/channels/573028991527550986/1194065428041957567/1194067213964030074):

> pick up admin store item edit stuff? it's pretty important since v1's doesn't work anymore bc multiple store gallery images

[discord](https://discord.com/channels/573028991527550986/1012204566122598431/1197388023059451924):

> basically since we have multiple store gallery images now, the old image admin page on v1 won't work anymore since it was built to support just one image
> 
> so the plan for this ticket is to write an interface for v2 (you can decide the design for now or go off of v1 if you want) that lets us manage items now since v1 won't be sufficient
> 
> the big difference is now we can upload multiple images and arrange the orders of them

SheepTester:
# Info

Closes #123. Depends on #94 

I added store item and collection edit and create pages under `/store/item/new`, `/store/item/[uuid]/edit`, `/store/collection/new`, and `/store/collection/[uuid]/edit`. Their code is based on the event edit/create page. I also added edit and create buttons to the public storefront pages, but admins can hide these with an option in the admin panel that stores this preference in a cookie.

This also adds a `Draggable` component for reordering lists of elements. You just give it an array and a render function, and it'll make each element draggable. It works best when every element is the same size.

## Changes

In addition to the changes made above,

- I changed apiRequests/Responses from .d.ts -> .ts because it was obscuring some type errors/unknown types
- I enforced file sizes for profile pictures and store merch images
- The collection and item pages return a 404 error for missing items instead of redirecting to the store home page
- In "view all items" mode, the storefront collection sliders subtly link to their collection pages

Changes after first reviews:

- I also updated the rest of the codebase to prefer `reportError` for API errors and `cond ? jsx : null` instead of `cond && jsx`
- ~~The service worker now fetches from the network first instead of loading stale data from the cache~~
- Moved extra arguments of `withAccessType` into an options object because it became unclear what they mean

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots

Edit buttons on various parts of the website

Storefront: collections view | Storefront: all items view | Item page
--- | --- |---
image|image|image

New collection page:
image

Edit item page:
image

Reordering images and options:

image
image


SheepTester:
image

farisashai (changes requested):
When you click "View store as member" on the store home page, can we trigger a refresh so that it actually switches to the user preview because right now nothing will change until refreshing. The alternative is store client state in the component which works too but is a little more overhead

farisashai (changes requested):
when the checkbox "Preview store as member" is checked, can you update the text in the button "Manage Store Merchandise" to "View Merch Store" so the actions of the checkbox and button below it are synchronized with the actual behavior that will take place

farisashai (changes requested):
i'm going to leave some comments here that are only relevant for the store item page too and review it all at once

- when you enter an item page with variants, the first option that isn't out of stock should be selected by default to prevent layout shift from the price appearing after selecting an option

farisashai (changes requested):
if you flip the "preview as member" checkbox -> go to store -> return to admin page and repeat this cycle while never refreshing the page, i'm seeing that the checkbox value can easily end up out of sync with the cookie value - i wonder if this is because the page is using outdated props and not re-rendering?

farisashai (changes requested):
idk how feasible this is but after you click "Create Collection" and get re-routed to the modify collection page, clicking back on your browser should ideally take you to the store home page instead of the create collection page because the form has already been submitted and shouldn't be created again

farisashai (changes requested):
on the "Create Store Item" page, what does quantity mean in the options? i think you might want to indicate what these fields are setting a little bit better. also it looks like you're missing a place to define variants such as S, M, L, XL as options

farisashai (changes requested):
when a collection or store item is created or edited successfully, i think you should take the user to the collection or item page where they can see their changes visually to receive confirmation that everything was set up right - i don't think someone would immediately want to modify the collection/item on the same form immediately after creating it (unless there is a mistake which they can see by viewing the item)

farisashai (changes requested):
might be nice to list which collection this item is a part of on the item page if that data exists easily

farisashai (changes requested):
you should enforce validation that the lifetime limit >= monthly limit

src/components/admin/store/CollectionDetailsForm/index.tsx
```diff
@@ -0,0 +1,182 @@
+import EventDetailsFormItem from '@/components/admin/event/EventDetailsFormItem';
```
farisashai: can you import this as a named import or does that introduce a cyclical dependency

src/components/admin/store/CollectionDetailsForm/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+import { StoreAPI } from '@/lib/api';
+import { MerchCollection } from '@/lib/types/apiRequests';
+import { PublicMerchCollection } from '@/lib/types/apiResponses';
+import { reportError } from '@/lib/utils';
+import Link from 'next/link';
+import { useRouter } from 'next/navigation';
+import { useState } from 'react';
+import { SubmitHandler, useForm } from 'react-hook-form';
+import { BsArrowRight } from 'react-icons/bs';
+import style from '../ItemDetailsForm/style.module.scss';
```
farisashai: seeking to learn why we're importing a css file outside the current folder

src/components/admin/store/CollectionDetailsForm/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+  defaultData?: Partial<PublicMerchCollection>;
+  token: string;
+}
+
+const CollectionDetailsForm = ({ mode, defaultData = {}, token }: IProps) => {
+  const router = useRouter();
+  const initialValues: FormValues = {
+    title: defaultData.title ?? '',
+    // ACM blue
+    themeColorHex: defaultData.themeColorHex ?? '#62b0ff',
```
farisashai: wondering if there's a way to pull the css var directly from a sass sheet for this

other option is just referencing a css variable if that works 
```suggestion
    themeColorHex: defaultData.themeColorHex ?? 'var(whatever-blue-variable)',
```

src/components/admin/store/CollectionDetailsForm/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+  } = useForm({ defaultValues: initialValues });
+
+  const [loading, setLoading] = useState<boolean>(false);
+
+  const resetForm = () => reset(initialValues);
+
+  const createCollection: SubmitHandler<FormValues> = async formData => {
+    setLoading(true);
+
+    try {
```
farisashai: can you move any client-side API calls to a StoreAdminManager file so we can isolate out any business logic relating to making the api call in that function and keep any stateful component updates in the callbacks to make it more readable and separate concerns like if the api schema changes for example we know where to look

src/components/admin/store/CollectionDetailsForm/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+      setLoading(false);
+    }
+  };
+
+  const editCollection: SubmitHandler<FormValues> = async formData => {
+    setLoading(true);
+
+    const uuid = defaultData.uuid ?? '';
+
+    try {
```
farisashai: same as above, move to a manager function

src/components/store/CollectionSlider/index.tsx
```diff
@@ -1,32 +1,60 @@
 import { Carousel } from '@/components/common';
+import CreateButton from '@/components/store/CreateButton';
```
farisashai, line 2: import all three of these from @/components/store in one line as named imports

SheepTester:
Cookies seem very finicky; I'll have to do some more testing with them (unsure how client vs server cookies work, and when they're synched). Originally preview mode was kept in a URL parameter rather than as a local preference

> idk how feasible this is but after you click "Create Collection" and get re-routed to the modify collection page, clicking back on your browser should ideally take you to the store home page instead of the create collection page because the form has already been submitted and shouldn't be created again

Oh this is real easy, I can just use router replace instead of push



> on the "Create Store Item" page, what does quantity mean in the options? i think you might want to indicate what these fields are setting a little bit better. also it looks like you're missing a place to define variants such as S, M, L, XL as options

I can change quantity to Quantity Available

The variant is the leftmost column that appears when there are two or more options

src/components/store/ItemCard/index.tsx
```diff
@@ -26,25 +27,34 @@ type ItemCardProps = (StoreItemOptions | CollectionOptions) & CommonOptions;
...
-      </div>
-    </Card>
+    <article className={`${styles.itemCard} ${className}`}>
+      <Link href={href} className={styles.linkWrapper}>
+        <div className={styles.imageWrapper}>
+          <Image src={image} alt={title} fill />
+        </div>
+        <div className={styles.details}>
+          <p className={styles.title}>{title}</p>
+          {'description' in props && <p>{props.description}</p>}
```
farisashai: im not 100% sure but i feel like it might be better to write this like `<p>{props?.description ?? ""}</p>` so react only has to update the text and we don't have the entire element rendering from scratch

src/components/store/SizeSelector/index.tsx
```diff
@@ -11,21 +11,23 @@ interface SizeSelectorProps {
 const SizeSelector = ({ currSize, options, onSizeChange }: SizeSelectorProps) => {
   const myID = useId();
```
farisashai, line 15: this might be a bit late but can someone explain to me why we have to define IDs and can't rely on component state to track everything it needs

src/pages/profile/edit.tsx
```diff
@@ -15,25 +15,14 @@ import withAccessType from '@/lib/hoc/withAccessType';
...
+import { capitalize, fixUrl, getProfilePicture, reportError } from '@/lib/utils';
 import DownloadIcon from '@/public/assets/icons/download-icon.svg';
 import DropdownIcon from '@/public/assets/icons/dropdown-arrow-1.svg';
 import styles from '@/styles/pages/profile/edit.module.scss';
-import { AxiosError } from 'axios';
 import type { GetServerSideProps } from 'next';
 import Link from 'next/link';
 import { FormEvent, useEffect, useId, useMemo, useState } from 'react';
 
-function reportError(title: string, error: unknown) {
```
farisashai, line 27: nice improvement! if you could confirm everywhere is using this function, that would be really good

src/pages/profile/edit.tsx
```diff
@@ -15,25 +15,14 @@ import withAccessType from '@/lib/hoc/withAccessType';
 import { CookieService, PermissionService } from '@/lib/services';
 import { PrivateProfile } from '@/lib/types/apiResponses';
 import { CookieType, SocialMediaType } from '@/lib/types/enums';
-import { capitalize, fixUrl, getMessagesFromError, getProfilePicture } from '@/lib/utils';
+import { capitalize, fixUrl, getProfilePicture, reportError } from '@/lib/utils';
 import DownloadIcon from '@/public/assets/icons/download-icon.svg';
 import DropdownIcon from '@/public/assets/icons/dropdown-arrow-1.svg';
 import styles from '@/styles/pages/profile/edit.module.scss';
```
farisashai:
```suggestion
import styles from '@/styles/pages/EditProfile.module.scss';
```
This isn't caused by your PR but can we flatten all of the page level CSS sheets to one folder by renaming them and fixing the imports because there's not really any reason to nest them in folders like we do with pages to define the routing structure.

src/pages/store/collection/[uuid].tsx
```diff
@@ -11,35 +11,53 @@ import styles from '@/styles/pages/StoreCollectionPage.module.scss';
...
 }: CollectionProps) => {
+  const canManageStore = PermissionService.canEditMerchItems.includes(accessType) && !previewPublic;
+
   return (
     <div className={styles.container}>
       <Navbar balance={credits} showBack />
       <div className={styles.header}>
         <Typography variant="h1/bold" component="h1">
           {title}
+          {canManageStore && <EditButton type="collection" uuid={uuid} />}
```
Comment (farisashai) [inline src/pages/store/collection/[uuid].tsx]: i might have missed this syntax existing in a few previous PRs but we actually have a standard on this for conditional rendering where we should always write it as 
 ```suggestion
          {canManageStore ? <EditButton type="collection" uuid={uuid} /> : null}
```

the bug this is avoiding is where you have a check such as 
```
const quantity = 0;
{quantity && <MyComponent />}
```
you would expect this not to render because the conditional is falsy however javascript shortcircuiting will still return the first value in this case so react will render a plaintext 0 on the screen. by writing it with the suggested syntax, this bug is not possible since the conditional is never the returned value

File: src/pages/store/index.tsx [hunk 12]
```diff
@@ -1,5 +1,14 @@
...
+import {
+  CollectionSlider,
+  EditButton,
+  HelpModal,
+  Hero,
+  HiddenIcon,
+  ItemCard,
+  Navbar,
+} from '@/components/store';
+import CreateButton from '@/components/store/CreateButton';
```
farisashai: import in group above

farisashai (changes requested):
excellent work mr sheeptester you are a frontend god and win the trophy for fastest and largest prs in history with a very clean implementation 🏆 i'm gonna be honest i stopped reading every line at some point but i think i did a decent code quality review - hopefully someone else can do a full run through of the diff. either way, for a feature of this size, we should have at least 3 different people stress test the preview deployment for edge cases and UI bugs that can be addressed before merging it so i would hold off on merging until we have enough validation in that sense.

two more comments i didn't mention yet
- i'm seeing a small bug on iOS when you drag an item image in edit UI where the box shadow leaves a trail across the page, i can try to screen record it later if you can't reproduce it
- on the admin actions homepage, the buttons text overflows on small screen sizes, if you can unhardcode the height? on these, or fix that UI issue, that would be excellent

src/lib/types/enums.ts
```diff
@@ -1,6 +1,7 @@
 export enum CookieType {
   USER = 'USER',
   ACCESS_TOKEN = 'ACCESS_TOKEN',
+  PREVIEW = 'PREVIEW',
```
farisashai: this variable isn't very descriptive currently as to what kind of browser state it's storing, we should rename to something like "USER_PREVIEW_ENABELD" or similar

src/components/admin/store/CollectionDetailsForm/index.tsx
SheepTester: They have the same CSS, so I didn't want to copy it. Ideally I would rearrange the files, but I'm not sure how. I'll find a way

src/components/store/ItemCard/index.tsx
SheepTester: I would rather the empty paragraph element not be there. Also, `description` for `ItemCard` shouldn't change often (and the way it's currently used, it won't change without a page reload)

src/components/store/SizeSelector/index.tsx
SheepTester: This comes from Andy's PR, but IDs are used so that inputs and labels can be flattened (`<input id={id} /> <label htmlFor={id}>...</label>` instead of `<label><input /></label>`). This allows the CSS to do `input:checked + label` etc and make `label` (which is not a replaced element so it's easier to style) to be what the user sees as the radio button, whilst still preserving built-in browser functionality.

I suppose we could use `label:has(:checked)` now, but that CSS feature feels uncomfortably new

We can use JavaScript to track the checked state of each radio button, but it's a lot more code than just relying on CSS.

src/components/admin/store/CollectionDetailsForm/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+        )}
+      </div>
+
+      <div className={style.form}>
+        <label htmlFor="title">Title</label>
+        <EventDetailsFormItem error={errors.title?.message}>
+          <input
+            type="text"
+            id="title"
+            placeholder="ACM Cafe"
```
alexzhang1618: nit: let's change the placeholder to something more relevant

src/components/admin/store/CollectionDetailsForm/index.tsx
alexzhang1618: Let's refactor `EventDetailsFormItem` into a common component named `DetailsFormItem` or `FormItem` since the file seems pretty generic and it's being used in two places now

src/components/admin/store/CollectionDetailsForm/index.tsx
alexzhang1618: Here and in ItemDetailsForm

src/components/admin/store/ItemDetailsForm/index.tsx
```diff
@@ -0,0 +1,571 @@
...
+                        { price: '', quantity: '', discountPercentage: '0', value: '' },
+                      ])
+                    }
+                  >
+                    Add option
+                  </Button>
+                </td>
+              </tr>
+            </tfoot>
+          </table>
```
alexzhang1618: Is this worth moving into a separate file to make the code more concise?

src/components/store/ItemCard/index.tsx
```diff
@@ -26,25 +27,34 @@ type ItemCardProps = (StoreItemOptions | CollectionOptions) & CommonOptions;
...
+          {'cost' in props && (
+            <p className={styles.cost}>
+              <Diamonds count={props.cost} />
```
alexzhang1618: Can we also add the discount value here on Diamonds? I've noticed it's not being displayed when it should

src/pages/admin/index.tsx
```diff
@@ -26,14 +28,27 @@ const AdminPage = ({ user: { accessType } }: AdminProps) => {
...
       <Typography variant="h2/bold">Store</Typography>
+      <label>
+        <input
+          type="checkbox"
+          defaultChecked={preview === 'member'}
+          onChange={e =>
+            CookieService.setClientCookie(
+              CookieType.PREVIEW,
+              e.currentTarget.checked ? 'member' : 'admin'
+            )
```
alexzhang1618: seeking to learn: should we set the client cookie or the server cookie here? I see in gSSP that we're picking from the server cookie

src/pages/store/collection/[uuid].tsx
```diff
@@ -11,35 +11,53 @@ import styles from '@/styles/pages/StoreCollectionPage.module.scss';
...
 
 const CollectionsPage = ({
-  user: { credits },
-  collection: { title, description, items = [] },
+  uuid,
+  user: { credits, accessType },
+  collection: { title, description, items = [], archived },
+  previewPublic,
 }: CollectionProps) => {
+  const canManageStore = PermissionService.canEditMerchItems.includes(accessType) && !previewPublic;
```
Comment (alexzhang1618) [inline src/pages/store/collection/[uuid].tsx]: nit: let's rewrite this into something more descriptive like storeAdminVisible

src/pages/store/index.tsx
```diff
@@ -9,18 +18,42 @@ import { getDefaultMerchItemPhoto } from '@/lib/utils';
...
+  user: { credits, accessType },
+  view,
+  collections,
+  previewPublic,
+}: HomePageProps) => {
+  const router = useRouter();
+
   const [helpOpen, setHelpOpen] = useState(false);
 
+  const canManageStore = PermissionService.canEditMerchItems.includes(accessType) && !previewPublic;
```
alexzhang1618: again, rewrite to something like canViewStoreAdmin?

src/pages/store/index.tsx
```diff
@@ -31,37 +64,71 @@ const StoreHomePage = ({ user: { credits }, view, collections }: HomePageProps)
       <div className={styles.container}>
         <div className={styles.header}>
           <h2>{view === 'collections' ? 'Browse our collections' : 'Browse all items'}</h2>
+          {canManageStore && (
+            <button
+              type="button"
+              className={styles.viewToggle}
+              onClick={() => {
+                CookieService.setClientCookie(CookieType.PREVIEW, 'member');
```
alexzhang1618: Same question here with client vs. server cookies

alexzhang1618 (commented):
Functionality is 🔥! 
There's some weird cookie stuff going on where with setting the view in settings and moving to the store doesn't reveal the new view unless you refresh. 
Do you think it'll work better if we use exclusively client cookies to check if we should display admin tools?

src/components/store/CollectionSlider/index.tsx
SheepTester: since this is also a store component, there will be a dependency circle because `@/components/store` already depends on `@/components/CollectionSlider`

src/pages/admin/index.tsx
SheepTester: looks like faris answered this on discord

image

SheepTester:
I messed with the next-pwa settings and now the admin preview cookie desync issues are gone :crab:

next.config.js
```diff
@@ -1,8 +1,26 @@
...
+// will always be up-to-date.
+// https://github.com/vercel/next.js/discussions/52024#discussioncomment-6325542
+const nextData = runtimeCaching.find(entry => entry.options.cacheName === 'next-data');
+if (nextData) {
+  nextData.handler = 'NetworkFirst';
+}
+
 const withPWA = require('next-pwa')({
   dest: 'public',
+  runtimeCaching,
```
alexzhang1618, line 23: Can we try merging main in and disabling caching for specific pages instead of site-wide?

It should just be 
`withAccessType(... true)` at the bottom of the page where gSSP is being exported

src/components/admin/store/DetailsForm/CollectionDetailsForm.tsx
```diff
@@ -0,0 +1,172 @@
...
+  } = useForm({ defaultValues: initialValues });
+
+  const [loading, setLoading] = useState<boolean>(false);
+
+  const resetForm = () => reset(initialValues);
+
+  const createCollection: SubmitHandler<FormValues> = async formData => {
+    setLoading(true);
+
+    const uuid = await AdminStoreManager.createNewCollection(token, formData);
```
farisashai: how are you error handling here if an api error gets thrown?

src/components/admin/store/DetailsForm/CollectionDetailsForm.tsx
```diff
@@ -0,0 +1,172 @@
...
+      setLoading(false);
+    }
+  };
+
+  const editCollection: SubmitHandler<FormValues> = async formData => {
+    setLoading(true);
+
+    const uuid = defaultData.uuid ?? '';
+
+    if (await AdminStoreManager.editCollection(token, uuid, formData)) {
```
farisashai: can we do async operations on their own line and just save it to a variable to use here for better readability

src/components/admin/store/DetailsForm/ItemDetailsForm.tsx
```diff
@@ -0,0 +1,361 @@
...
+          setPhoto(null);
+          if (reason === 'cannot-compress') {
+            showToast(
+              'Your image has too much detail and cannot be compressed.',
+              'Try shrinking your image.'
+            );
+          } else if (reason !== null) {
+            showToast('This image format is not supported.');
+          }
+        }}
```
alexzhang1618: It feels like we generally send the same error for every Cropper component, thoughts on bringing this inside the component and not allowing it to be customized?

alexzhang1618 (approved):
we move 🚢

next.config.js
SheepTester: I don't think `disableCaching` works. Clicking on "View store as member" doesn't do anything until you reload the page or click again, like it was originally.

farisashai (approved):
small comments but looks good

PR #125, opened 2024-01-19
Better cookies

farisashai:
refactor user cookie validation logic to include a retry mechanism to gracefully re-generate the user cookie if it's corrupted without forcing the user to log in again


SheepTester (approved):
lgtm

PR #126, opened 2024-01-19
even better user cookies

farisashai:
# Info

Implemented improvement suggested here: https://discordapp.com/channels/573028991527550986/1012204566122598431/1197469917071548436

Since cookies do not store a date created attribute, the idea is that whenever the user cookie is set on the client, it is given a maxAge of 5 minutes, after which it is automatically deleted on the browser. Since the auth token has a lifespan of 2 weeks, it can be used repeatedly to re-generate the user cookie whenever it expires and deletes. (When the auth token expires, we are forced to logout to re-generate a new one).

This solves the problem of requesting an up-to-date version of the user object on every single page that needs it because now they will all rely on the shared cookie and this cookie itself will be updated whenever it is needed and does not exist.

This improvement relies on the assumption that any client side API modifications to the user object will also update it in real-time such as the way the edit profile functions so it is never out of sync with the API on a single device.

The only edge case that will fail is checking into an event or editing a user profile on 2 separate devices in under 5 minutes, which may display a invalid UI for this duration until it get's wiped and updated to the correct data subsequently.

With this change, the user API will be hit once every 5 minutes at most instead of on every page route change.

## Changes
- fix a small bug where eventmodal wasn't using the fixUrl function to pad links correctly
- implement feature described above

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots


src/lib/api/UserAPI.ts
```diff
@@ -40,6 +42,25 @@ export const getCurrentUser = async (token: string): Promise<PrivateProfile> =>
   return response.data.user;
 };
 
+export const getCurrentUserAndRefresh = async (
```
alexzhang1618: Can we name this into something like `getCurrentUserAndRefreshCookie`? Right now it's not obvious what 'refresh' means

src/lib/managers/AccountManager.ts
alexzhang1618, line 1: Are we getting rid of the error wrapping around a user call with a manager? 

I think we should keep it so we can fail gracefully if the cookie is invalid/the API call fails and we can redirect to the homepage.

One weird bug that I'll get when working locally is that if I switch from the prod API to the testing API or vice versa, it'll try to fetch a user based on my cookie and error out since the uuid doesn't exist, which results in a ugly Vercel 500 error screen instead of redirecting to login like it probably should. I'd then have to manually clear cookies to fix this.

I'd like it if we could wrap the API call in a manager and clear the cookie if the API call fails/the user isn't found.

alexzhang1618 (approved):
Changes generally LGTM, just some questions:

src/lib/managers/AccountManager.ts
farisashai: This is in the actual API file so any usage of this function with a manager function should remain unaffected. I didn't error handle for a corrupted cookie in here because that's already handled in withAccessType that gets called beforehand so we just need to check whether the cookie exists or not

alexzhang1618: Then can we migrate the `getCurrentUserAndRefresh` calls made in other pages (index, profile, etc) to manager-wrapped calls? 
The error still happens on this branch since I don't think the cookie checking in `withAccessType` checks if the user exists:
Screenshot 2024-01-19 at 9 12 25 AM

farisashai: You're only hitting this error because the validation passes since the user cookie is completely valid and not-corrupt, however the auth token is invalid when you switch base API urls. this isn't something that's possible in any real scenarios so I think it's fine to leave as it and just clear your cookies if you have to switch between prod and testing on localhost

farisashai: I don't know if I've ever clarified but the convention established was to just use manager functions for client side API calls to provide a clean way to separate API logic from component state logic with the callbacks. There isn't really a point here, if we do want to resolve the issue, we can just wrap the actual API call in a try-catch

Issue #127, opened 2024-01-20
EventModal Improvements


farisashai:
![image](https://github.com/acmucsd/membership-portal-ui-v2/assets/33165426/cd2df094-6a20-4d99-b8b7-471f636ab15f)

- [x] Remove the add to calendar buttons for past events (and currently happening?)
- [x] Fix whitespace wrapping
- [x] Indicate attendance in the modal and prompt to give feedback if they haven't already submitted anything for an attended event

PR #138, opened 2024-01-27
Event Modal tweaks

alexzhang1618:
# Info

Closes **#127**.

## Changes

- Add whitespace breaking to event modal description
- Rework mobile view of event modals to be more visually appealing
- Remove calendar buttons if an event is currently/has already happened

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots

https://github.com/acmucsd/membership-portal-ui-v2/assets/27360825/02f60ce7-7818-4925-b43e-de8ffef9fd3f


farisashai:
on mobile when the content is really long, there are two types of scrolling happening. If you drag inside the modal, you'll scroll the content inside and if you drag outside the modal on the backdrop, you'll scroll the entire modal itself. Can you disable internal scrolling and just make the modal take all the height it requires?

https://github.com/acmucsd/membership-portal-ui-v2/assets/33165426/42cb3519-5334-430b-9234-47da95a3a56b

farisashai:
I would also decrease the modal padding on mobile significantly, the content is insanely narrow

src/components/common/Modal/index.tsx
```diff
@@ -7,15 +7,42 @@ interface ModalProps {
...
 
   useEffect(() => {
+    /**
+     * In order to stop the body behind the modal from scrolling in iOS,
+     * we need to add position: fixed and overflow: hidden to body and html (this is done in globals.scss)
+     *
+     * A side effect of this is that the position is reset to the top of the page every time the modal
+     * resets. We address this by saving and restoring the scroll position every time we close a modal.
+     */
+    const body = document.querySelector('body');
```
SheepTester: you can just use `document.body`

src/pages/_document.tsx
```diff
@@ -7,6 +7,10 @@ export default function Document() {
         <link rel="manifest" href="/manifest.json" />
         <link rel="apple-touch-icon" href="/apple-touch-icon.png" />
         <meta name="theme-color" content="#fff" />
+        <meta
+          name="viewport"
+          content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0"
+        />
```
SheepTester: This currently prevents mobile users from zooming in, which is pretty bad for accessibility. You only need `initial-scale`.

```suggestion
        <meta
          name="viewport"
          content="width=device-width, initial-scale=1.0"
        />
```

If the mobile browser is zooming the page out regardless, then there's probably some horizontal scroll due to an element that is too wide, which should be fixed instead of disabling zoom

src/styles/globals.scss
```diff
@@ -9,6 +9,12 @@ body {
   scroll-padding-top: 4.25rem;
 }
 
+/* stylelint-disable-next-line selector-no-qualifying-type */
+body:has(dialog[open]) {
+  overflow: hidden;
+  position: fixed;
```
SheepTester: this causes a shift when the dialog opens (the scroll position gets sent back to the top)

I think we could just put off fixing scrolling behind dialogs and blame it on browsers instead. Maybe it'll be fixed in a few years (https://github.com/whatwg/html/issues/7732)

src/components/common/Modal/index.tsx
```diff
@@ -7,15 +7,42 @@ interface ModalProps {
   open: boolean;
   onClose: () => void;
   children?: ReactNode;
+  // Anchor content to the bottom of the screen. Primarily for mobile modals.
```
SheepTester: I think you can change this to a doc comment

```suggestion
  /** Anchor content to the bottom of the screen. Primarily for mobile modals. */
```

so in VS Code it will show this description in autocomplete

SheepTester (changes requested):
The gap at the top still goes away when I scroll down (Chrome, both desktop and Android), unlike in the video

image

Also, I think you should still have an explicit close or back button. You can just copy the HTML for the close button. No JS is required; it just has to be a submit button

```tsx
<button type="submit" aria-label="Close">
  <CloseIcon aria-hidden />
</button>
```

src/pages/_document.tsx
alexzhang1618: Just going to remove this tag since I don't think it really accomplished anything functionally

alexzhang1618:
> The gap at the top still goes away when I scroll down (Chrome, both desktop and Android), unlike in the video

This change is intended and I think it looks better/more consistent with other modals I've seen. I've also added a close button for the modal and I'll update the video clip accordingly

src/components/common/Modal/index.tsx
```diff
@@ -7,9 +7,11 @@ interface ModalProps {
   open: boolean;
   onClose: () => void;
   children?: ReactNode;
+  /** Anchor content to the bottom of the screen. Primarily for mobile modals. */
```
raymosun: It looks like bottomSheet doesn't apply unless you're on mobile now. Maybe you can update the comment to "Only applies for mobile modals"

raymosun (commented):
I really like the new mobile look! 

One weird thing is the blue carousel scrollbar gets inherited to the modal scrollbar now 🤔. Also, do you think the modal starts out too big? It's a little strange to open the modal and only see the same information from the card until you scroll down.

image
(Desktop Chrome)

src/components/common/Modal/style.module.scss
```diff
@@ -37,12 +37,19 @@
     background: none;
   }
 
+  @media screen and (width <= vars.$breakpoint-md) {
+    &.bottomSheet {
+      padding: 4rem 0 0;
```
SheepTester, line 43: image

should also remove the bottom border radii when it's in bottom sheet mode

src/styles/globals.scss
```diff
@@ -9,6 +9,12 @@ body {
   scroll-padding-top: 4.25rem;
 }
 
+/* This prevents the background from scrolling while a modal is open. */
+/* stylelint-disable-next-line selector-no-qualifying-type */
+body:has(dialog[open]) {
+  overflow: hidden;
```
SheepTester, line 15: low priority but hiding the scrollbar on windows causes a layout shift when the dialog opens


https://github.com/acmucsd/membership-portal-ui-v2/assets/22133785/e5453136-8f7c-4255-8300-01821d3debf9

it's probably more trouble than it's worth trying to fix it, though. one idea is to do

```css
html {
  overflow-y: scroll; /* force scrollbar to take up space */
}
body {
  display: fixed; /* disable scrolling */
}
```

but `overflow-y: scroll;` will make a scrollbar that takes up space appear on Macs even when it normally wouldn't

SheepTester (approved):
do you think the close button would look nicer as a circle and with an opaque 

image image

PR #140, opened 2024-01-27
First experiments with user banners:

alexzhang1618:
# Info

Closes **#120**.

- Still in progress

## Changes

- [Fill in here]

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [ ] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots

### DIFF
--- a/src/components/profile/UserProfilePage/index.tsx
+++ b/src/components/profile/UserProfilePage/index.tsx
@@ -1,6 +1,7 @@
 import { Carousel, GifSafeImage, Typography } from '@/components/common';
 import { EventCard } from '@/components/events';
 import SocialMediaIcon from '@/components/profile/SocialMediaIcon';
+import UserBanner from '@/components/profile/UserBanner';
 import { config, showToast } from '@/lib';
 import { PublicAttendance, type PublicProfile } from '@/lib/types/apiResponses';
 import { SocialMediaType } from '@/lib/types/enums';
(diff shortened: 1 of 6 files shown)


src/components/profile/UserProfilePage/index.tsx
```diff
@@ -34,7 +35,7 @@ export const UserProfilePage = ({
   return (
     <div className={styles.profilePage}>
       <div className={styles.cardWrapper}>
-        <div className={styles.banner} />
+        <UserBanner attendances={isSignedInUser ? signedInAttendances : attendances} />
```
farisashai, line 38: if you view raymond's banner, it's blank because his attendances are private. maybe have some fallback

farisashai:
right now every user's profile banner looks exactly the same. I think you should play around with using a random seed of the user's uuid to generate the location of the colors or something like that, and then you can calculate the intensity using their attendances or default to an equal intensity for all communities if the attendances are private.

src/components/profile/UserProfilePage/index.tsx
farisashai: also since we want this feature to be a public part of the user profile, we can't rely on the user's own attendance either if it's private because that will display differently for them than how other users see their profile

PR #116, opened 2024-01-07
Create Store Cart Page

raymosun:
# Info

Closes #57 

# Description

Implement the store cart page. Currently using regular events as placeholder pickup events until the link to events from pickup events is merged on backend.

## Changes

- Creates the Store Cart page at /store/cart

# Type of Change

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [ x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as
      expected)
- [ ] Logistics Change (A change to a README, description, or dev workflow setup like
      linting/formatting)
- [ ] Continuous Integration Change (Related to deployment steps or continuous integration
      workflows)
- [ ] Other: (Fill In) 

# Testing

I have tested that my changes fully resolve the linked issue ...

- [ x ] locally on Desktop.
- [ x ] locally on mobile - use https://ngrok.io to get a copy on a mobile device
- [ ] on the live deployment preview on Desktop.
- [ ] on the live deployment preview on Mobile.
- [ ] I have run and passed all new and existing Cypress tests. Add screenshots below.

# Checklist

- [ ] I have performed a self-review of my own code.
- [ ] I have followed the style guidelines of this project.
- [ ] I have documented my code's `src/lib` functions and commented hard to understand areas
      anywhere else.
- [ ] My changes produce no new warnings.

# Screenshots

*Need to add more but need to set up a way to simulate a full cart

image

PR #141, opened 2024-01-29
Implement cart page (2)

### LINKED ISSUE #57: Create Store Cart Page
Implement store cart page at `/store/cart` based on these [figma designs](https://www.figma.com/file/cgbeW6vrqgxW0i1vpbUn1C/Store-(Redesign-2023)?type=design&node-id=654-2155&mode=design&t=jCa9Nm6GIE583hHg-0) (check the 'ready for dev' sections). 

Requires #54.

Encompasses:
- Desktop layout for cart page
- Mobile layout for cart page
- Cart Item component
- Pickup confirmation modal

image

raymosun:
# Info

Closes #57. The [old branch](https://github.com/acmucsd/membership-portal-ui-v2/pull/116)'s linting check got out of sync (possibly because the branch is very old), and the changes are being redone here.

While using the testing API, you can add some items to your cart cookie by navigating to `/debug` and clicking the 'Fill cart' button.

## Changes

- Creates the Store Cart page at `/store/cart
- Some imports of UUID were updated to use `import type`
- Adds the `/debug` page, visible when using the testing API, to house debug functionalities like adding items to the cart

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots

image
image


src/components/store/CartItemCard/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const itemPage = `${config.itemRoute}${item.uuid}`;
+
+  return (
+    <div className={styles.cartItem}>
+      {item.merchPhotos?.[0] && (
+        <div className={styles.leftCol}>
+          <Link href={itemPage}>
+            <div className={styles.imageWrapper}>
+              <Image
+                src={item.merchPhotos[0].uploadedPhoto}
```
alexzhang1618: We have `getDefaultMerchItemPhoto` in utils.ts to get the default merch photo instead of doing `item.merchPhotos?.[0]`

I also think using this will remove the need to do a check if there is a merchPhoto, so you can just write the div instead of doing an `&&`

src/components/store/CartItemCard/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+          )}
+          <Typography variant="h5/bold" component="p">
+            Quantity:&nbsp;
+            <Typography variant="h5/regular" component="span">
+              {item.quantity}
+            </Typography>
+          </Typography>
+        </div>
+      </div>
+      {removable && (
```
alexzhang1618: What does `removable` mean? In what case is an item not removable?

src/components/store/StoreModal/index.tsx
```diff
@@ -0,0 +1,77 @@
+import { Modal, Typography } from '@/components/common';
+import { PropsWithChildren, ReactElement, cloneElement, useState } from 'react';
+import styles from './style.module.scss';
+
+interface StoreModalProps {
+  opener: ReactElement<any>;
+  title: string;
+  onConfirm?: (...args: any[]) => any;
+  onCancel?: (...args: any[]) => any;
+}
```
alexzhang1618: Do we have a better name than `StoreModal`? I think we have other modals in the store (like Sean's store info modal). Maybe something like `StoreConfirmationModal` might make more sense?

src/lib/types/index.d.ts
alexzhang1618, line 1: Why is this being changed?

src/pages/store/cart.tsx
```diff
@@ -0,0 +1,336 @@
...
+                quantity: 2,
+              },
+              {
+                itemUUID: 'afe7cfff-9a62-4146-8c8c-5b14f39ba0a0',
+                optionUUID: '72d22f34-c366-4ecd-a365-059661a0f0fc',
+                quantity: 1,
+              },
+            ];
+            setClientCookie(CookieType.CART, JSON.stringify(newCart));
+            showToast('Filled your cart, buddy-o', 'Refresh to apply changes');
```
alexzhang1618: can we do `buddy-o!`

alexzhang1618 (commented):
Haven't really gotten a chance to play around with the UI, but here are some code nits I had. Also the transform biggening on the linked event thingy for cart I think we should fix

src/components/store/CartItemCard/index.tsx
raymosun: I pop up the item card again inside of the modal for removing and I implemented this using recursion, the removable prop is like a base case that stops it having another card inside itself.

src/components/common/Modal/index.tsx
farisashai, line 1: seeking to learn: why did this need to be changed?

src/pages/store/cart.tsx
```diff
@@ -0,0 +1,336 @@
...
+  let savedItems;
+  try {
+    const cartCookie = getServerCookie(CookieType.CART, { req, res });
+    savedItems = JSON.parse(cartCookie);
+    if (!Array.isArray(savedItems)) throw new Error();
+  } catch {
+    savedItems = [];
+  }
+
+  const savedCartPromises = Promise.all(
```
farisashai, line 275: i think this is worth opening an issue with backend to request a new api endpoint to fetch a set of store items by giving a list of uuids since it's a since database read on their end and doesn't have to open a new network connection for each item. the functionality already exists 100% so it's an easy addition of returning the data formatted differently

farisashai, line 275:
```suggestion
  const savedCart = Promise.all(
```
nit to rename the variable because technically this isn't returning a promise array since you chained the `.then()` on it

src/pages/store/cart.tsx
```diff
@@ -0,0 +1,336 @@
...
+      } catch {
+        return undefined;
+      }
+    })
+  ).then((items): ClientCartItem[] => items.filter(item => item !== undefined) as ClientCartItem[]);
+
+  const pickupEventsPromise = getFutureOrderPickupEvents(AUTH_TOKEN).then(events =>
+    events.filter(event => event.linkedEvent && event.status !== 'CANCELLED')
+  );
+  const userPromise = getCurrentUser(AUTH_TOKEN);
```
farisashai: i introduced a slightly tweaked version of this function that performs the same functionality but refreshes the user cookie at the same time that we should be using again, you might need to pull main if you don't have that code existing

src/pages/store/cart.tsx
```diff
@@ -0,0 +1,336 @@
...
+
+  // update the cart cookie if part of it was malformed
+  setServerCookie(CookieType.CART, clientCartToCookie(savedCart), { req, res });
+
+  return {
+    props: {
+      user,
+      savedCart,
+      pickupEvents,
+      token: AUTH_TOKEN,
```
farisashai: since the auth token exists in a cookie, we can fetch it client side when we need it to avoid prop drilling and reduce the size of the serversideprops return value which affects how long the page load takes

src/pages/store/cart.tsx
```diff
@@ -0,0 +1,336 @@
...
+  }, [cart, credits]);
+
+  // prop for item cards to remove their item
+  const removeItem = (optionUUID: UUID) => {
+    setCart(cart => cart.filter(item => item.option.uuid !== optionUUID));
+  };
+
+  // handle confirming an order
+  const placeOrder = () =>
+    placeMerchOrder(token, {
```
farisashai: this should work but I started using a specific convention in this repo of wrapping client-side api calls in a manager function that would be in the src/lib/managers/StoreManager. If we follow this pattern, it'll keep the code a little bit cleaner by separating business logic required to interact with the api and any logic that is only relevant for modifying the component UI state. also the client side function can easily be moved to a different component file and we only need to tweak the UI effects that are relevant for the new component in the callback function

src/pages/store/cart.tsx
```diff
@@ -0,0 +1,336 @@
...
+  JSON.stringify(
+    items.map(item => ({
+      itemUUID: item.uuid,
+      optionUUID: item.option.uuid,
+      quantity: item.quantity,
+    }))
+  );
+
+const StoreCartPage = ({ user: { credits }, savedCart, pickupEvents, token }: CartPageProps) => {
+  const calculateOrderTotal = (cart: ClientCartItem[]): number => {
```
farisashai: it doesn't look like this function depends on the component props or state besides the arguments it takes in so I think we can move it outside this component so it gets defined exactly once in the JS bundle

src/components/store/StoreModal/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const [open, setOpen] = useState(false);
+
+  const onClose = () => {
+    setOpen(false);
+    onCancel();
+  };
+
+  return (
+    <>
+      {cloneElement(opener, {
```
farisashai: seeking to learn: what is this?

farisashai (dismissed):
some initial code quality feedback - i dont know how i would add anything to the cart rn so maybe i'll look at the preview after we merge andy's PR

src/components/store/StoreModal/index.tsx
raymosun: this was my solution for attaching an onclick to the item that gets passed as the opener (usually a button that will open the modal), idk if there's a better solution for this

src/pages/store/cart.tsx
raymosun: hey, to clarify, which functionalities would be moved to the StoreManager file:
- forming the `{ orders, pickupEvent }` object?
- passing the token?
- showing the toast?

src/pages/store/cart.tsx
raymosun: Doesn't `.then()` return another promise? My understanding is since nothing has been awaited, no promises have been resolved, and API requests haven't been made until savedCartPromises is awaited with the others below.

farisashai: Oop yea I think you're right

src/components/common/Modal/index.tsx
raymosun: this was refactored to allow modals inside modals, for the event card in the confirmation modal; there was a hydration error caused by the browser not allowing `<form>` inside `<form>`

src/pages/store/cart.tsx
raymosun: you got it buddy-o!

raymosun:
Updated page to handle sold-out/limit-reached/hidden items in latest commits

src/pages/store/cart.tsx
raymosun: So true, there was actually a route that was supposed to do this but it it's broken

src/components/store/StoreConfirmModal/style.module.scss
```diff
@@ -0,0 +1,54 @@
...
+      max-width: 12.5rem;
+    }
+
+    .confirm {
+      background: vars.$success-1;
+    }
+
+    .goBack {
+      background: vars.$danger-1;
+    }
```
SheepTester: the colors on this

image

are confusing. I would expect red / neutral colors because you're confirming a dangerous action

src/components/store/StoreConfirmModal/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+                }}
+                className={styles.confirm}
+              >
+                <Typography variant="h4/medium" component="span">
+                  Confirm
+                </Typography>
+              </button>
+              <button type="button" onClick={onClose} className={styles.goBack}>
+                <Typography variant="h4/medium" component="span">
+                  Go back
```
SheepTester: would use "Cancel" because it's a modal; "go back" implies going to the previous page (leaving the cart page)

src/lib/utils.ts
```diff
@@ -298,3 +303,28 @@ export const fixUrl = (input: string, prefix?: string): string => {
...
+/**
+ * Check if a ClientCartItem is in stock and within lifetime/monthly limits
+ * @param item item to validate
+ * @returns an error message string if the item is unavailable, or null otherwise
+ */
+export const validateClientCartItem = (item: ClientCartItem): string | null => {
+  if (item.quantity > item.lifetimeRemaining)
+    return `Cannot purchase more than the lifetime limit of this item (${item.lifetimeRemaining} left)`;
+  if (item.quantity > item.monthlyRemaining)
+    return `Cannot purchase more than the monthly limit of this item (${item.monthlyRemaining} left)`;
```
SheepTester: I feel like the error messages could be better worded, like maybe

> You have reached your lifetime limit for this item.
>
> You have reached your monthly limit for this item.

if remaining is 0, and

> You may only purchase up to {remaining} more of this item.
>
> You may only purchase up to {remaining} more of this item for this month.

if remaining > 0, or something

src/components/store/PickupEventPicker/style.module.scss
```diff
@@ -0,0 +1,57 @@
...
+      svg {
+        color: var(--theme-text-on-background-3);
+        transition: color 0.25s ease;
+      }
+
+      &:disabled {
+        cursor: default;
+
+        svg {
+          color: transparent;
```
SheepTester, line 53: image

in the design, it looks like the arrow becomes an outline when disabled. which looks better, do you think?

image

src/components/store/CartItemCard/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  return (
+    <div className={styles.cartItem}>
+      <div className={styles.leftCol}>
+        <Link href={itemPage}>
+          <div className={styles.imageWrapper}>
+            <Image
+              src={getDefaultMerchItemPhoto(item)}
+              alt={item.itemName}
+              sizes={`(max-width: ${config.cssVars.breakpointMd}) 100px, 150px`}
```
SheepTester, line 54: out of curiosity, what does this do?

src/components/store/CartItemCard/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+      </div>
+      <div className={styles.rightCol}>
+        <div className={styles.cartItemInfo}>
+          <div className={styles.title}>
+            <Link href={itemPage}>
+              <Typography variant="h4/medium" component="h3">
+                {item.itemName}
+              </Typography>
+            </Link>
+            <Diamonds count={item.option.price * item.quantity ?? 0} className={styles.price} />
```
SheepTester: is the `?? 0` necessary? I feel like TypeScript wouldn't let you multiply them if one of them was `number | undefined`

src/components/store/PickupEventPicker/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const PickupEventPicker = ({ events, eventIndex, setEventIndex, active }: EventPickerProps) => {
+  return (
+    <div className={styles.eventPicker}>
+      <div className={styles.window}>
+        {events.length > 0 ? (
+          <div
+            className={styles.slider}
+            style={{ transform: `translateX(${-320 * eventIndex}px)` }}
```
SheepTester: Apparently you can pull variables from SCSS. For example,

```scss
$eventWidth: 320px;

.eventPicker {
  .window {
    width: $eventWidth;
  }
}

:export {
  #{eventWidth}: $eventWidth;
}
```
```ts
import styles from './style.module.scss'

const WIDTH = parseFloat(styles.eventWidth)
```

Maybe that'd be nice here in case we want to change the width later

SheepTester: You could also just use a CSS variable:

```scss
.eventPicker {
  .window {
    --width: 320px;
    width: var(--width);
  }
}
```

```tsx
<div style={{ transform: `translateX(calc(var(--width) * ${eventIndex}))` }}>
```

or

```scss
.eventPicker {
  .window {
    width: 320px;
    
    .slider {
      transform: translateX(calc(-320px * var(--index)));
    }
  }
}
```

```tsx
<div style={{ '--index': eventIndex }}>
```

but this one might have a type error in TypeScript

src/components/store/StoreConfirmModal/style.module.scss
```diff
@@ -0,0 +1,54 @@
...
+  }
+}
```
SheepTester: should end the file with an extra newline

src/lib/managers/StoreManager.ts
```diff
@@ -0,0 +1,31 @@
...
+    StoreAPI.placeMerchOrder(token, {
+      order: items.map(({ option: { uuid }, quantity }) => ({
+        option: uuid,
+        quantity,
+      })),
+      pickupEvent,
+    })
+      .then(res => resolve(res))
+      .catch(err => reject(err.response.data.error));
+  });
```
SheepTester: You should just convert this to an async function

```suggestion
export const placeMerchOrder = async (items: ClientCartItem[], pickupEvent: UUID) =>
  const token = getClientCookie(CookieType.ACCESS_TOKEN);
  if (!token) throw new Error('Missing access token');

  return StoreAPI.placeMerchOrder(token, {
    order: items.map(({ option: { uuid }, quantity }) => ({
      option: uuid,
      quantity,
    })),
    pickupEvent,
  });
}
```

then use `reportError` from lib/utils to handle the error (since `err.response.data.error` might itself cause an error if there was a network error rather than api error)

src/components/events/EventCard/style.module.scss
```diff
@@ -10,6 +8,11 @@
   transition: all 0.3s;
   width: 20rem;
 
+  &.bordered { 
+    border: 1px solid var(--theme-primary-6);
+    border-radius: 10px;
+  }
+
   &:hover {
```
SheepTester: prevent the button from having a hover effect if clicking it doesn't do anything

ideally, you should also make it a `<div>` or smth rather than a `<button>` , like

```tsx
const children = <>...</>;

return <>
  {isOrderPickupEvent(event) && !event.linkedEvent ? <div>{children}</div> : <button type="button" onClick={...}>{children}</button>}
</>;
```

src/components/events/EventDetail/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+import LinkIcon from '@/public/assets/icons/link.svg';
+import Image from 'next/image';
+import Link from 'next/link';
+import styles from './style.module.scss';
+
+interface EventDetailProps {
+  event: PublicEvent | PublicOrderPickupEvent;
+  attended: boolean;
+  /** controls whether a close button is shown */
+  inModal?: boolean;
```
SheepTester, line 16: Should this be named `showClose` then?

src/components/events/EventDetail/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+          {isUpcomingEvent && !isOrderPickupEvent(event) ? <CalendarButtons event={event} /> : null}
+        </div>
+
+        <Typography variant="body/medium" style={{ wordBreak: 'break-word' }}>
+          {description}
+        </Typography>
+        <Link className={styles.link} href={displayEventLink}>
+          <div style={{ width: 11 }}>
+            <LinkIcon role="link" />
```
SheepTester: I think `role="link"` is redundant here because `Link` (an `<a>` element) already has a link role. you should probably use `aria-label` or `aria-hidden` instead

src/components/store/CartItemCard/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+      opener={
+        <button type="button" className={styles.removeBtn}>
+          <Typography variant="h5/regular" component="span">
+            Remove Item
+          </Typography>
+        </button>
+      }
+      title="Are you sure you want to remove this item?"
+      onConfirm={() => {
+        setTimeout(() => removeItem && removeItem(item.option.uuid), 100);
```
SheepTester, line 38: you should add a comment explaining why there's a delay

src/pages/store/cart.tsx
```diff
@@ -0,0 +1,323 @@
...
+  const placeOrder = () =>
+    placeMerchOrder(cart, pickupEvents[pickupIndex]?.uuid ?? 'x')
+      .then(({ items, pickupEvent }) => {
+        showToast(
+          `Order with ${items.length} item${items.length === 1 ? '' : 's'} placed`,
+          `Pick up at ${pickupEvent.title}`
+        );
+        setCartState(CartState.CONFIRMED);
+        setClientCookie(CookieType.CART, clientCartToCookie([]));
+        setLiveCredits(credits - orderTotal);
```
SheepTester, line 88: You should also set the `USER` cookie with the new credit count

src/pages/store/cart.tsx
```diff
@@ -0,0 +1,323 @@
...
+          <div />
+        )}
+
+        {/* Confirmation */}
+        {cartState === CartState.CONFIRMED ? (
+          <div className={`${styles.cartCard} ${styles.confirmation}`}>
+            <Typography variant="h3/regular" component="p">
+              You will receive a confirmation email with your order details soon.
+              <br />
+              Please pick up your items on the date you selected. Thank you!
```
SheepTester: should it say

```suggestion
              Please pick up your items at the event you selected. Thank you!
```

instead? unsure

src/pages/store/cart.tsx
```diff
@@ -0,0 +1,323 @@
...
+                  </Typography>
+                </button>
+              </Link>
+              <Link href={config.myOrdersRoute}>
+                <button type="button" className={styles.storeButton}>
+                  <Typography variant="h5/bold" component="span">
+                    My Orders
+                  </Typography>
+                </button>
+              </Link>
```
SheepTester: don't put `<button>`s instead links

src/pages/store/cart.tsx
```diff
@@ -0,0 +1,323 @@
...
+              <Typography variant="h6/regular" component="p">
+                Your cart is empty. Visit the ACM Store to add items!
+              </Typography>
+              <Link href={config.storeRoute}>
+                <button type="button" className={styles.storeButton}>
+                  <Typography variant="h5/medium" component="span">
+                    Go to Store
+                  </Typography>
+                </button>
+              </Link>
```
SheepTester: same here

src/pages/store/cart.tsx
```diff
@@ -0,0 +1,323 @@
...
+
+        // form the ClientCartItem from the PublicMerchItem
+        const ClientCartItem: Partial<ClientCartItem> = { ...publicItem, quantity };
+        ClientCartItem.option = options.find(
+          (option: PublicMerchItemOption) => option.uuid === optionUUID
+        );
+
+        // check for invalid cart cookie item
+        if (!ClientCartItem.option || typeof quantity !== 'number') throw new Error();
+        return ClientCartItem as ClientCartItem;
```
SheepTester: I think you should make the variable name lowercase so this line doesn't look as confusing

SheepTester (approved):
lgbtm, a bunch of nitpicks below

also, you should pull in changes from main because #94 got merged

src/pages/store/cart.tsx
raymosun: I like this 👍

src/lib/managers/StoreManager.ts
raymosun: the new getErrorMessage function is awesome!

src/components/store/CartItemCard/index.tsx
raymosun: This lets the next-image component know what resolution to give the image in even though it's responsive https://nextjs.org/docs/pages/api-reference/components/image#sizes

src/components/store/PickupEventPicker/index.tsx
raymosun: thanks for the tip!

src/components/store/PickupEventPicker/style.module.scss
raymosun: i kinda like the no arrow tbh

src/pages/store/cart.tsx
raymosun: I think the USER cookie is refreshed whenever the user visits a page that displays their credit count. I didn't set it on this page and had the credit balance calculator still use the old cookie state.

src/pages/store/cart.tsx
raymosun: I refactored to a manager in newer versions 👍

PR #179, opened 2024-03-25
Redesign dashboard

alexzhang1618:
# Info

## Changes

- Redesign the dashboard to match elevation + with design tweaks

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots

Screenshot 2024-03-25 at 4 27 11 PM
Screenshot 2024-03-25 at 4 27 38 PM


SheepTester:
Not a fan of the inconsistent spacing here

Screenshot_20240325-171533

src/components/home/HomeActions/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  checkin: (code: string) => void;
+}
+
+const HomeActions = ({ user, points, checkin }: HomeActionsProps) => {
+  const [checkinCode, setCheckinCode] = useState('');
+
+  const handleSubmit = (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    if (checkinCode !== '') {
+      checkin(checkinCode);
```
SheepTester, line 20: should it reset the text in the check-in code input when you successfully check in

src/components/profile/UserProgress/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+export const UserProgress = ({
+  user,
+  points,
+  isSignedInUser,
+  levelTextVariant,
+  levelDescriptionVariant,
+}: UserProgressProps) => {
+  const { firstName } = user;
+  // animate the progress bar
+  const [progress, setProgress] = useState<Number>(0);
```
SheepTester:
```suggestion
  const [progress, setProgress] = useState<number>(0);
```

src/components/profile/UserProgress/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  useEffect(() => setProgress(points % 100), [points]);
```
SheepTester, line 25: This could probably also be done with a CSS animation with a CSS variable, which would avoid needing a re-render

src/components/profile/UserProgress/index.tsx
SheepTester, line 1: Should we move calculations of stuff like `points % 100` and `100 - points % 100` to a separate file/function so it's easier to change ranks to a logarithmic scale in the future?

src/components/profile/UserProgress/index.tsx
```diff
@@ -0,0 +1,62 @@
+import { Typography } from '@/components/common';
+import { Variant } from '@/components/common/Typography';
```
SheepTester:
```suggestion
import { type Variant } from '@/components/common/Typography';
```

I think

SheepTester: also I think `@/components/common` should also re-export `Variant` so you can merge the first two imports

SheepTester (approved):
You could reduce the amount of duplicate elements and desktop-/mobile-only styles by using a CSS grid

For example, a desktop `grid-template-areas` could look like

```css
grid-template-areas:
  'welcome welcome'
  'checkin events'
  'level   events'
  'history history';
```

where the hero image spans behind `welcome`. Then on mobile, you could keep the same elements but change the template areas to

```css
grid-template-areas:
  'welcome'
  'checkin'
  'level'
  'events'
  'history';
```

with the hero spanning across the first three rows

Also, since we got rid of `liveEvents` I wonder if it would be nice to have an `ONGOING` badge for events that are currently ongoing

src/components/home/HomeActions/index.tsx
alexzhang1618: Sure

src/components/profile/UserProgress/index.tsx
alexzhang1618: Hmm we can do this later down the line since it's only being used in this file

src/components/profile/UserProgress/index.tsx
alexzhang1618: Not an expert on this but I think one nice thing from this is that the points animation will jump by 10 right after a successful check-in; can we also do this with CSS?

SheepTester: you can use a CSS transition. something like this might work

```scss
@keyframes points {
  from {
    width: 0;
  }
  to {
    width: var(--progress);
  }
}
.inner {
  animation: points .5s;
  width: var(--progress);
  transition: width .2s;
}
```

then you can just set the `--progress` style to a percentage

PR #199, opened 2024-03-30
Read query params on initial load for events page

### LINKED ISSUE #187: Events Page Should Accept Query Params
If we add support for URL query parameters, we can improve the links in the event card carousels so they launch the page in a matching view that was shown in the carousel with the initial state of the filter options.

Example:
Clicking "See Events" from this carousel should go to the link `/events?attended=true&range=pastyear` so it sets the page to only show attended events in the past year.
image

alexzhang1618:
# Info

Closes #187 .

## Changes

- Read query params on initial load for events page
- Updating some event carousel links so we can use query params
  - viewing another user's attendances is more involved (maybe future pr)
- Note: Experimented with using query params as a state but updating the browser's URL was laggy, so I left it out
  - Maybe worth using a library for this?

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.
- [x] I have added new Cypress tests that are passing.

# Checklist

      anywhere else.

# Screenshots


src/pages/events.tsx
```diff
@@ -58,22 +59,72 @@ const filterEvent = (
...
+    const attendance = searchParams.get('attendance');
+    if (AttendanceOptions.some(({ value }) => value === attendance)) {
+      setAttendedFilter(attendance as string);
+    }
+
+    const query = searchParams.get('query');
+    if (query !== null) {
+      setQuery(query);
+    }
+  }, [searchParams, years]);
```
SheepTester: I'm getting a layout shift from this because it reads the URL parameters in a `useEffect`

Since `searchParams` and `years` both won't change after loading the page, I think you could try something like

```ts
const { defaultCommunityFilter, defaultTimeFilter, defaultAttendedFilter, defaultQuery } = useMemo(() => getDefaultFilters(searchParams, years), [searchParams, years]);
const [communityFilter, setCommunityFilter] = useState(defaultCommunityFilter);
const [timeFilter, setTimeFilter] = useState(defaultTimeFilter);
const [attendedFilter, setAttendedFilter] = useState(defaultAttendedFilter);
const [query, setQuery] = useState(defaultQuery);
```

SheepTester (commented):
For some reason. clicking "Events" in the navbar doesn't reset the filters on the page:

1. Go to home page
2. Click on "See all events." This filters events by "Attended"
3. Click on "Events" in the navbar. Nothing on the page changes, but the URL updates to `/events`

> Note: Experimented with using query params as a state but updating the browser's URL was laggy, so I left it out
> * Maybe worth using a library for this?

What kind of lag are you experiencing? I think browsers will add some delay if you change the URL too often (like multiple times per second) to avoid crashing the browser

In my experience, using `window.history.replaceState` is usually pretty fast, but I'm not sure how compatible it is with Next.js's router. [this stackoverflow post](https://stackoverflow.com/a/77956963) says to use `router.replace` instead, with `shallow: true` to avoid reloading

it would be great to have that feature, though, so it doesn't get reset when you leave and come back, and it improves discoverability of the feature

farisashai:
I tried a different approach in https://github.com/acmucsd/membership-portal-ui-v2/pull/202 by handling query logic server-side to determine the initial value of the page state and leaving the client alone fully. I think it should solve the layout shift and route change issues we're facing here and be easier to maintain in the future.

farisashai (approved):
🚢 🚢 🚢

PR #218, opened 2024-04-02
Hotfix: Fix users not being able to submit a second feedback

SheepTester:
# Info

Currently, users can only submit one feedback in their lifetimes, since the rest of the events will list that one feedback they submitted instead of allowing them to submit feedback for that event.

This is because currently, the event page uses query parameters to filter feedback by user and event ID. However, these query parameters are ignored if the user doesn't have the permission to see all feedback.

I think this is an issue with the backend's design, but whatever. Instead, this now fetches all feedback (for admins, this will fetch every feedback ever submitted) and then filters through that array. This matches the current behavior of checking for attendance: to check if a user has attended a specific event, it fetches the user's entire attendance history then looks up the event ID.

Reproduction steps:

1. Create a normal user account
2. Give feedback for one past event
3. Attempt to give feedback for another past event. Currently, the page will show the feedback given in #2 on the second event's page. This PR fixes the issue and will allow the user to give feedback for the second event.

## Changes

- Fetch all feedback and filter the array by user and event ID, rather than relying the backend's filters

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots


SheepTester:
fixed by acmucsd/membership-portal#424

PR #228, opened 2024-04-14
Make events not require sign-in

SheepTester:
# Info

there's no issue for this

event pages now no longer require login (because the API never required it). this way, you can send links to the events in Discord and it'll show the event cover bigly

however, to encourage people to sign up, these pages have a massive banner telling people to log in. the login form is embedded so you can just click log in if you have a password saved

there are otherwise no changes to the user's experience when signed in

## Changes

- event page doesn't require log in
  - also added preview images for several pages, and the event page's preview image is LARGE on discord
- events page doesn't require login
- login appeal component that includes the login form that shows when the user is logged out
  - useful if/when store becomes readonly for non-signed in users too
- i also refactored all the pages so that `withAccessType` passes it the auth token (since it already reads it from the cookie)

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots

image

image

image

image

i also fixed the blue (for some reason Discord parses `#fff` as `#00f`, oof)

image


farisashai:
image

I think this might be too big on mobile since there's essentially no relevant content to the events page above the fold. Maybe expandable login section with just an alert showing or smth smaller in general

farisashai:
there's no way to get to the single
event page from the events page modals to share a link to them ?

farisashai:
If you click sign up from the events page login popup and then go back, it takes you from the /register page to the login page instead of the page you were previously on. I think this might have been hardcoded in initially

farisashai:
You should hide the attended/not attended filter if the events page isn't logged in

SheepTester:
> If you click sign up from the events page login popup and then go back, it takes you from the /register page to the login page instead of the page you were previously on. I think this might have been hardcoded in initially

I'm not able to reproduce this for individual event pages

1. Go to event page, e.g. https://membership-portal-ui-v2-git-sean-public-event-acmucsd.vercel.app/events/e154acf7-3465-49e4-baa6-f40bfa2f4ad8
2. Go back
3. It takes me back to the event page

For /events, going to "Sign up" and back leaves me on the /register page but changes the URL to /events. I think this is due to a bug in the event's page's filtering implementation rather than the registration page



> there's no way to get to the single event page from the events page modals to share a link to them ?

I would like to add "Events" to the navbar even when signed out. I think I'll do this in a separate PR after the store page also becomes public

I think the current navbar implementation isn't great (has duplicate elements for mobile and desktop; has a hardcoded height; uses signed-out navbar on 404 pages) so I'd like to refactor it in a separate PR, along with adding events and store for the logged out navbar



> [image](https://private-user-images.githubusercontent.com/33165426/322295532-3d46fb6e-553e-4232-9280-089774a780f1.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTMzOTk1OTMsIm5iZiI6MTcxMzM5OTI5MywicGF0aCI6Ii8zMzE2NTQyNi8zMjIyOTU1MzItM2Q0NmZiNmUtNTUzZS00MjMyLTkyODAtMDg5Nzc0YTc4MGYxLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA0MTglMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNDE4VDAwMTQ1M1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWJjY2MzMzU3MTYwOTUxODg2MDE1MWY5MTNkOTJlNDFhMGJmNzNmMjc1MDc5YWE5YmVkYmZmMzg3YmE2ODNjYjcmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.WMqTW2gW4w-LQ967VuiAVD-y4oUKSj_G7gdHzykex3A)
> 
> I think this might be too big on mobile since there's essentially no relevant content to the events page above the fold. Maybe expandable login section with just an alert showing or smth smaller in general

I think the size is fine. We'd rather people stay logged in so we should try to pester them to log in, and it's easy to scroll past anyways

src/components/common/Login/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+          required: 'Required',
+        })}
+        error={errors.password}
+      />
+      <SignInButton
+        type="link"
+        display="link"
+        text="Forgot your password?"
+        href="/forgot-password"
+      />
```
alexzhang1618: Should we migrate this to something like `FormButton` or `FormItem`?

src/lib/api/EventAPI.ts
```diff
@@ -20,10 +20,11 @@ import axios from 'axios';
 /**
  * Get a single event by UUID
  * @param uuid Search query uuid
- * @param token Bearer token
+ * @param token Bearer token. Optional, but should be provided if you need to
+ * see attendance codes.
  * @returns Event info
  */
-export const getEvent = async (uuid: UUID, token: string): Promise<PublicEvent> => {
+export const getEvent = async (uuid: UUID, token?: string): Promise<PublicEvent> => {
```
alexzhang1618, line 27: Is it worth just removing token and updating everywhere if it's not needed?

alexzhang1618 (approved):
Approving so this isn't blocked, but some questions

src/lib/api/EventAPI.ts
SheepTester: I was going to, but it turns out the token affects whether you can see the access code for events, so removing it would break the admin event page

PR #244, opened 2024-04-30
fulfill

### LINKED ISSUE #225: Admin - Fulfill Pickup Event Items

SheepTester:
# Info

Closes #225 

fulfills request for fulfill button

## Changes

- fulfill button for each order
- checkboxes for items that can be deselected to mark as not fulfilled
- got rid of fulfill view

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots

image

image


src/components/admin/store/PickupOrder/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+              try {
+                const items = order.items.filter(item => !unselected.has(item.uuid));
+                const newOrder = await StoreAPI.fulfillOrderPickup(token, order.uuid, items);
+                const itemUuids = items.map(item => item.uuid);
+                onOrderUpdate({
+                  ...newOrder,
+                  items: order.items.map(item =>
+                    itemUuids.includes(item.uuid) ? { ...item, fulfilled: true } : item
+                  ),
+                });
```
alexzhang1618: seeking to learn: does the API response not just return the whole new order updated correctly? do we need to manually set fulfilled for the items we fulfilled

alexzhang1618 (commented):
Code looks good! Good call on removing the order fulfill view i was just parroting off of v1 even though we don't really need it

Some usability comments:

can we put some gap between the status and fulfill button and maybe make the button the same size as the fulfill tag
Screenshot 2024-05-10 at 3 37 34 PM

also i think it would be nice to sort by status (i.e. put orders that still need action at the top)

I think we discussed this but I think ux wise it makes sense for all checkboxes to start blank and the merch person will tick each box off as they get the item

Another thing to consider: is it better to put the fulfill button under the merch items for each user? I think it's nicer so your mouse only has to stay on one side but idk

src/components/admin/store/PickupOrder/index.tsx
SheepTester: the API response doesn't return the `items` array so it won't say which items were fulfilled

SheepTester:
I changed the orders view to use badges instead. not sure if it's better

image


Marcelo cancelled this rescheduled partially fulfilled order, which is why it has one fulfilled item:

image

src/components/admin/store/PickupOrder/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+        <ul className={styles.itemList}>
+          {itemQuantities.map(item => {
+            let badge = null;
+            if (
+              order.status === OrderStatus.FULFILLED ||
+              order.status === OrderStatus.PARTIALLY_FULFILLED
+            ) {
+              if (!item.fulfilled) {
+                badge = <span className={styles.notFulfilled}>Not fulfilled</span>;
+              }
```
alexzhang1618:
```suggestion
            if (
              !item.fulfilled && (
                order.status === OrderStatus.FULFILLED || 
                order.status === OrderStatus.PARTIALLY_FULFILLED
              )
            ) {
                badge = <span className={styles.notFulfilled}>Not fulfilled</span>;
```

src/components/store/OrderSummary/index.tsx
```diff
@@ -57,6 +49,12 @@ const OrderItemPreview = ({ item, orderStatus }: OrderItemPreviewProps) => {
           alt="Store item picture"
           fill
         />
+        {showFulfilled && item.fulfilled ? (
+          <div className={styles.badge}>{showNotFulfilled ? 'Picked up' : 'Already picked up'}</div>
+        ) : null}
+        {showNotFulfilled && !item.fulfilled ? (
+          <div className={`${styles.badge} ${styles.notAvailable}`}>Not available</div>
```
alexzhang1618: Not available makes me think that it's not in stock, can we do something like not picked up maybe? Could also ask others on what wording is best

src/components/store/OrderSummary/index.tsx
```diff
@@ -57,6 +49,12 @@ const OrderItemPreview = ({ item, orderStatus }: OrderItemPreviewProps) => {
           alt="Store item picture"
           fill
         />
+        {showFulfilled && item.fulfilled ? (
+          <div className={styles.badge}>{showNotFulfilled ? 'Picked up' : 'Already picked up'}</div>
```
alexzhang1618: Thoughts on like a green for this instead of grey similar to how notAvailable is red?

alexzhang1618 (commented):
I like the tags and the extra text, it looks nice!
Some UI comments:

I think I prefer ✅ and ❌ emojis to the text since it's more concise and readable

I'm also not sure how much I like having all these check boxes to the right of the item. In our store, we don't really have a lot of items where people are ordering a big quantity of them, but maybe we should make this just one check to fulfill all or maybe a number to fulfill a specific number of them? Up to you and whatever looks best
Screenshot 2024-05-18 at 12 26 44 PM

src/components/store/OrderSummary/index.tsx
SheepTester: yeah it should only show for partially fulfilled events, so it wasn't in stock

image

SheepTester:
- I switched the text labels in the admin orders page back to emoji
- I merged all the checkboxes into one checkbox per variant. This means that if someone orders multiple ACM AI stickers, but there's somehow only one sticker left, you can only either mark them all as fulfilled or not fulfilled. Maybe that's fine for now.
- I changed the "picked up" / "already picked up" (for rescheduled orders) badge to green:
  
  image

src/components/admin/store/PickupOrder/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+                onOrderUpdate({
+                  ...newOrder,
+                  items: order.items.map(item =>
+                    itemUuids.includes(item.uuid) ? { ...item, fulfilled: true } : item
+                  ),
+                });
+              } catch (error: unknown) {
+                reportError('Failed to fulfill order', error);
+              }
+            }}
```
alexzhang1618: tiny nit but if there's a good amt of code here i'd rather it be a helper function `handleFulfillOrder` so we can keep the jsx nice and tidy

alexzhang1618 (approved):
LGTM, some nits

Maybe a word break and a little margin between the checkbox n the button?
Screenshot 2024-08-19 at 11 01 14 PM

SheepTester:
image

PR #246, opened 2024-05-02
Filter out pickup events that start in less than 2 days

### LINKED ISSUE #241: Pickup Event Hide Pickup Events Less Than 2 Days Ahead
Right now, users are shown pickup events that are less than two days ahead. They can't select these pickup events as an option, so we should hide them.

image

nik-dange:
# Info

Closes #241

Filtered out pickup events that start less than 2 days from now, and hides them from the Cart page. There were some discussions around if they should be hidden or just prevent from being selected but ultimately we decided hiding.

## Changes

- filter

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots

### DIFF
--- a/src/pages/store/cart.tsx
+++ b/src/pages/store/cart.tsx
@@ -254,6 +254,13 @@ const getServerSidePropsFunc: GetServerSideProps = async ({ req, res }) => {
       .filter(
         event => !(event.orders && event.orderLimit && event.orders.length > event.orderLimit)
       )
+      // filter out events that have a start time less than 2 days from now
+      .filter(event => {
+        const startTime = new Date(event.start);
+        const now = new Date();
+        now.setDate(now.getDate() + 2);
+        return startTime >= now;
+      })
   );
   const userPromise = getCurrentUserAndRefreshCookie(AUTH_TOKEN, { req, res });
 
--- a/src/pages/store/cart.tsx
+++ b/src/pages/store/cart.tsx
@@ -257,9 +257,9 @@ const getServerSidePropsFunc: GetServerSideProps = async ({ req, res }) => {
       // filter out events that have a start time less than 2 days from now
       .filter(event => {
         const startTime = new Date(event.start);
-        const now = new Date();
-        now.setDate(now.getDate() + 2);
-        return startTime >= now;
+        const now = Date.now();
+        const twoDaysFromNow = new Date(now + 2 * 24 * 60 * 60 * 1000);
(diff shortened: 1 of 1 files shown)


src/pages/store/cart.tsx
```diff
@@ -254,6 +254,13 @@ const getServerSidePropsFunc: GetServerSideProps = async ({ req, res }) => {
       .filter(
         event => !(event.orders && event.orderLimit && event.orders.length > event.orderLimit)
       )
+      // filter out events that have a start time less than 2 days from now
+      .filter(event => {
+        const startTime = new Date(event.start);
+        const now = new Date();
+        now.setDate(now.getDate() + 2);
```
SheepTester: Doing it this way might make it prone to edge cases when days are 23 or 25 hours long. I think it's fine, but it might not match the backend perfectly

I think a more precise implementation would compare the difference in milliseconds with 2 days' worth of milliseconds, rather than relying on how the date is represented in the user's time zone

SheepTester (approved):
Lgtm

src/pages/store/cart.tsx
WishingWell13: https://stackoverflow.com/questions/12795767/trying-to-add-3-days-in-milliseconds-to-current-date

maybe

SheepTester: yeah, could do something like `startTime.getTime() >= now.getTime() + PICKUP_ADVANCE_TIME` or something, where `PICKUP_ADVANCE_TIME = 1000 * 60 * 60 * 24 * 2`

nik-dange: See updated code, does that have the same effect? Lots of different Date methods and ways of representing them

src/pages/store/cart.tsx
```diff
@@ -257,9 +257,9 @@ const getServerSidePropsFunc: GetServerSideProps = async ({ req, res }) => {
       // filter out events that have a start time less than 2 days from now
       .filter(event => {
         const startTime = new Date(event.start);
-        const now = new Date();
-        now.setDate(now.getDate() + 2);
-        return startTime >= now;
+        const now = Date.now();
+        const twoDaysFromNow = new Date(now + 2 * 24 * 60 * 60 * 1000);
```
SheepTester: This is good, though I think you could move that number into a constant

SheepTester (approved):

PR #264, opened 2024-09-18
Use Pacific Time on server side

SheepTester:
# Info

image

4 am?? oops

Uses Pacific Time when displaying dates on server side only. The client continues to use local time.

## Changes

It seems the Vercel servers are in a different time zone (maybe UTC?), so when displaying dates on the server side, I changed it to use the America/Los_Angeles time zone. This fixes the times rendered in event web previews. The client uses local time, like before.

An alternative option would be to force Pacific Time on the client. This could make sense since ACM at UCSD is based in California. However, we still host online events like Leetcode sessions, and people might be attending them from outside of California, so it would be helpful to continue showing them the event's time in their time zone.

# Type of Change

      expected)
      linting/formatting)
      workflows)
- [ ] Other: (Fill In) 

# Testing

- [x] locally on Desktop.

# Checklist

      anywhere else.

# Screenshots

image

the event image says october 2 but i don't think that's my fault

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -183,19 +183,22 @@ export function useObjectUrl(file?: Blob | null): string {
   return url;
 }
 
+// Set timeZone to Pacific Time on server-side, use local time on client. Users
+// will usually be in California, so this helps with hydration.
 const dateFormat = new Intl.DateTimeFormat('en-US', {
   month: 'short',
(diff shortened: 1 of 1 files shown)


alexzhang1618 (approved):
LGTM, is it worth specifying that it's PST in these?

PR #279, opened 2026-05-20
added new s3 testing domain

edward-lemonade:
i renamed the s3 instances to be less misleading, between testing and prod versions. naming convention on aws is now consistent with hackathon-portal


NathanRLiu (approved):
lgtm

